PR #1, opened 2023-09-09
Make admin page load

bunnymatic:
Problem
----------

Because the api is busted and for other reasons, the admin dashboard (http://localhost:9292/admin) wouldn't load.

Solution
----------

This pr takes a few steps to make it work and add some more developer handy notes in the README.

Changes
--------

* update README
* add `racksh` gem so we can have a console
* add `rerun` gem so it's easier to develop and get live updates as you're working
* update admin/layout.html to render the dashboard template safely (move to `!=` from `=` in `admin/layout`
* trimmed some whitespace from pictures haml template


bunnymatic:
image

PR #8, opened 2023-09-16
Rubocop autocorrect most of lib/ dir

bunnymatic:
Problem
--------

We have a pile of rubocop todos that can be autocorrected.

So as not to cause huge merge issues, doing some selective updates.

Solution
--------

Remove most of `lib/*` files from rubocop_todo and run rubocop autocorrect.


bunnymatic (commented):
this did not increase test failures so seems like a move in the right direction.

PR #12, opened 2023-09-17
Kneyzberg/serialization cleanup

kneyzberg:
- Problem: Api from Mau changed, causing issues with Artists pages rendering. 

- Solution:
Fixed Artist and Artists model to generate serialized data correctly from Mau.


bunnymatic (approved):
Looks good.   we agreed to leave the tests busted and will be fixing them in the future.

👍

PR #16, opened 2023-09-17
Update Studio and Mediums models and reinstate tests

bunnymatic:
Problem
----------

We had a failing test around Studio and we had this `Mediums` model that was really bothering me because it reminded me of the `Artists` model which we did some rejiggering with in #13 and are working to make it look more like ActiveRecord style.

Solution
----------

* rerun studio vcr records and fix that Studio test
* remove most of the models/studio.rb rubocop violations
* rename Mediums to Medium
* add `.all` and `.find` methods to Medium
* add/update specs for Medium as appropriate
* clean up Rubocop violations around Medium class


models/medium.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  def name
+    @model['name']
+  end
+
+  def self.find(id)
+    all.find { |m| m.id == id }
+  end
+
+  def self.all
```
bunnymatic, line 18: @kneyzberg not sure if this is useful but it was a little reminiscent of the work we were talking about with `Artists` -> `Artist.all`.  You'll notice the `Mediums` deleted file has the same silly class instance variable and weird overuse of `media` variable name in it.  

When you add the artist find method, you might check here and see if it's helpful or not.

spec/models/artist_spec.rb
```diff
@@ -1,6 +1,10 @@
 require 'spec_helper'
 
 describe Artist, vcr: { record: :none } do
+  before do
```
bunnymatic, line 4: @kneyzberg also worth noting, I found that I had a few tests that were passing but when I put them on CI the did not.  This is because locally, I was actually caching the data so the tests weren't really starting clean slate.

We may find anywhere we use the cache in these model specs, we should do this `SafeCache.flush` before running any tests.

we might consider - do we want test that is "if the cache is there, it uses it" and "if not it fetches".  I'm not sure.  But something we can talk about.

models/medium.rb
kneyzberg: Sounds good, i'll take a look!

spec/models/artist_spec.rb
kneyzberg: Copy that, thanks for the info.

PR #20, opened 2023-09-23
Convert Contact page to slim

bunnymatic:
Problem
----------

we're using haml.  we like slim better.

solution
---------

Convert the `contact.haml` page to `slim` using `haml2slim` and a bunch of manual intervention because that script doesn't *really* to everything corretly.

Updated `breadcrumb.html` rendering to unescape it.


views/contact.slim
```diff
@@ -0,0 +1,20 @@
+.sidebar
+  h3 Contact
+  ul.contacts_list
+    li.contact
+      a href="http://maps.google.com?q=1890+Bryant+Street+San+Francisco,+CA"
+        .name google map
+.content
+  == haml :breadcrumb, :layout => false
```
bunnymatic, line 8: And in slim you need to add this `==` for the raw html rendering.

All this makes me thing we might rethink the way the `breadcrumb` is handled because it's kind of stupid

there's a template `breadcrumb.haml` that renders a div and calls a helper method `Breadcrumbs.new.render` which generates a `haml` string which we inject in the template.

Maybe the helper code should either render html or put the code in template so it's clearer what's going on .
And less escaping of stuff.

kneyzberg: I think that makes sense, definitely would want things to be more readable

kneyzberg (approved):

PR #21, opened 2023-09-23
Update ContentResource to not require 'section' and update specs

bunnymatic:
problem
-------

The initial migration made `ContentResource` which is our teeny CMS record to require a 'section'.  But looking at the tests and the code that we skipped, it's clear that we don't really care about section.  it was for future expansion.  So...

solution
--------

This PR adds a migration that relaxes the requirement on section for that model, and re-instates the code (and test) for the event page to actually show that content if it's in the system.


spec/models/content_resource_spec.rb
```diff
@@ -1,28 +1,31 @@
 # frozen_string_literal: true
 
 require 'spec_helper'
-require 'byebug'
```
bunnymatic, line 4: don't really need this here.

views/events.haml
```diff
@@ -4,4 +4,4 @@
   = haml :breadcrumb, :layout => false
   %h1 Events
   .markdown
-    = markdown_content(@content_body)
+    != markdown_content(@content_body)
```
bunnymatic, line 7: Another directive to `haml` to stop escaping html and just draw what I tell you!  :D

spec/app_spec.rb
```diff
@@ -88,21 +83,25 @@ def setup_fixtures
   end
 
   describe '#events' do
-    before do
-      # putting the get here doesn't seem to work
-      setup_fixtures
+    context 'when there is a content resource for events' do
+      before do
+        ContentResource.create({ page: 'events', body: "## Here's what we've got planned\n\n* this\n*that\n\n" })
```
bunnymatic, line 88: Moved this from the `setup_fixtures` helper to the context.  Now that we have ActiveRecord and `database_cleaner` the `setup_fixtures` which was cleaning records and adding them is no longer needed.


Note to self, think about factory bot?

PR #28, opened 2023-09-25
Dump HAML

bunnymatic:
problem
--------

trying to replace all haml with slim

solution
---------

the last place was in the breadcrumbs so we refactor the template to not use the `BreadCrumb` service but just do the logic and let the partial send in the breadcrumbs we want/need

changes
-------

* remove `haml`
* remove `BreadCrumb` service in favor of template logic


app.rb
```diff
@@ -3,24 +3,26 @@
 require 'bundler/setup'
 require 'sinatra'
 require 'sinatra/config_file'
+require 'sinatra/activerecord'
```
bunnymatic, line 6: just a little sorting/organizing

PR #35, opened 2023-09-30
 Change databases: pg -> mysql

bunnymatic:
problem
--------

You never really think you'll swap out databases but for our current
deployment, mysql is a little easier to deal with because it's already on those machines.

https://www.pivotaltracker.com/story/show/186838523

solution
--------

move to mysql for all environments cuz our linode doesn't have Postgres so...

Once this goes in, we'll have to rebuild databases on local machines with something like
```
bin/rake db:create db:migrate db:seed
RACK_ENV=test bin/rake db:create db:migrate db:seed
```
from there we should be back to normal.
And you could likely stop your Postgres server assuming you don't need it for other projects/work.

notes 
--------
You never really think you'll swap out databases but for our current deployment, mysql is a little easier to deal with because it's already on those machines.


db/migrate/20240113033938_update_studio_event_description_to_text.rb
```diff
@@ -0,0 +1,9 @@
+class UpdateStudioEventDescriptionToText < ActiveRecord::Migration[7.1]
+  def up
+    change_column :studio_events, :description, :text
```
bunnymatic, line 3: added because Postgres string fields allow more content and tests failed without this change.


ContentBlock was already using `text` so we didn't need that mod.

db/schema.rb
```diff
@@ -10,39 +10,36 @@
...
-  create_table "image_resources", force: :cascade do |t|
+  create_table "image_resources", charset: "utf8mb3", force: :cascade do |t|
     t.string "filename", null: false
     t.string "format", null: false
     t.string "mime_type", null: false
     t.integer "width", null: false
     t.integer "height", null: false
     t.bigint "file_size", null: false
-    t.binary "image", null: false
+    t.binary "image", size: :medium, null: false
```
bunnymatic, line 34: might be worth looking at what this means... do we need `size: :large`?  maybe better is move to s3 or some other cloud storage.

yarn.lock
```diff
@@ -1851,29 +1851,10 @@
     "@typescript-eslint/types" "5.62.0"
     "@typescript-eslint/visitor-keys" "5.62.0"
 
-"@typescript-eslint/scope-manager@6.18.0":
```
bunnymatic, line 1854: not sure what happened here.  this branch was really old so maybe just getting up to speed with all our yarn updates since I started this stuff

db/schema.rb
kneyzberg: I think if we get rid of that highlighted photo, we might not need large? Unless we replace original with large in the Carousel

kneyzberg (approved):

db/schema.rb
bunnymatic: This isn't about art piece or artist images.  This is storing images for the events.  it's a dumb way to do it but the `image` is a blob field in the db which means you can put binary data in there.  So when someone uploads an image for an event, this is where we stick it.

It was a simple way to not put it on the filesystem and not connect S3 but the *right* way is one of those two options.  And S3 is better cuz we can get some caching and CDN stuff.  I added a story to do that instead.  some day

PR #43, opened 2023-10-02
Add FakeCache client

bunnymatic:
problem
-------

in tests we ran into issues on CI where memcache was not available. we really rely on the cache to do our business.

solution
---------

add a `FakeCache` that acts like a real cache but is just in memory so that we can use it in tests.

And force it in tests with `spec/support/fake_cache.rb` so that all tests automagically get it.


lib/fake_cache.rb
```diff
@@ -0,0 +1,31 @@
...
+# something, please add to it
+# ***
+
+class FakeCache
+  def initialize
+    flush
+  end
+
+  def set(key, value, options = {})
+    @cache[key] = { value: value, options: options }
```
bunnymatic, line 20: we could some day honor the `expires_in` options in tests.  since we don't care yet, no need but this `value/options` syntax will support that if we cared.

lib/safe_cache.rb
```diff
@@ -1,13 +1,7 @@
...
 class SafeCache
   def self.cache
-    @cache ||= if ['MEMCACHIER_SERVERS', 'MEMCACHIER_USERNAME', 'MEMCACHIER_PASSWORD'].all? { |k| ENV.key? k }
-                 Dalli::Client.new(ENV['MEMCACHIER_SERVERS'].split(','),
-                                   { username: ENV.fetch('MEMCACHIER_USERNAME', nil),
-                                     password: ENV.fetch('MEMCACHIER_PASSWORD', nil) })
-               else
-                 Dalli::Client.new('localhost:11211', { expires_in: BryantStreetStudios.settings.cache_expiry })
-               end
+    @cache ||= new
```
bunnymatic, line 4: We do this so we can in tests - stub `SafeCache.new` to return a fake cache and then we get automatic injection of the fake in test env.

spec/lib/fake_cache_spec.rb
```diff
@@ -0,0 +1,26 @@
+require_relative '../../lib/fake_cache'
+
+RSpec.describe FakeCache do
+  subject(:cache) { described_class.new }
+
+  describe 'lifecycle behavior' do
```
bunnymatic, line 6: I looked at writing tests for each method but it didn't make sense not to see them all working together.

spec/support/fake_cache.rb
```diff
@@ -0,0 +1,5 @@
+RSpec.configure do |config|
+  config.before do
+    SafeCache.new(FakeCache.new)
```
bunnymatic, line 3: Make it so all tests receive this fake thingy.

kneyzberg (approved):
I like this, this seems to be working as expected, and not causing any issues. 

It makes sense to me that we would mock the cache for testing, it seems simpler and more straightforward that configuring the CI to have memcached.

PR #44, opened 2023-10-05
Surfaced Medium Id for ArtPiece Model and Wrote JsonApiHelper Class

kneyzberg:
#### Problem 
- currently, our ArtPiece model was not accessing the medium Id related to Art Pieces. 
- the data processing of JSON data in Artist and ArtPiece model was too verbose and clunky. 

### Solution
- surfaced medium id from json data and included it in Art Piece model.
- wrote JsonApiHelper class for processing json data, and wrote generalized `process_json_data()` and `find_medium_id()` method.
- refactored Artist and ArtPiece models to use JsonApiHelper.


lib/json_api_helpers.rb
```diff
@@ -0,0 +1,17 @@
...
+    formatted_data = {
+      'id' => data['id'].to_i,
+      **data['attributes']
+    }
+    formatted_data['artist_info'] = data['attributes']['artist_info']['artist_info'] if model == 'artist'
+    formatted_data['medium_id'] = find_medium_id(data) if model == 'art_piece'
+    formatted_data
+  end
+
+  def self.find_medium_id(data)
```
bunnymatic, line 12: I don't think these belong here.  `JsonApiHelper` should not know about all the model intricacies.  It should know about the commonalities of all jsonapi results.  It can know "there's a 'data' thing" and an "attributes" thing and a 'relationships' thing.  But it should know about medium_id or artist_info.  these things should live in their corresponding models that needs to know how special they are.  So the artist unpacker lives under `Artist` and the medium helper lives in `ArtPiece`

I could see the method here be something like
```ruby
class JsonApiHelper
  def self.process_json_data(data, relationships: []) 
     id = data['data']
     attributes = data['attributes']
     relationships = data['relationships']
     result = { id: id, **attributes }
     # pull out relationships that the user asked for in relationship
     relationship_items = data['relationships'].map{ |k,v| v if relationships.include?(k) }.compact
     relationship_items.each { |item| result.merge( item ) }

     result
  end
end
```

maybe that's it.  then maybe a helper in `artist` like

```
class Artist
   ...

   def unpack(raw_artist_data) 
     formatted = process_json_data(raw_artist_data)
     { artist_info: formatted.delete(artist_info), **artist_info }
  end

...
```
and the medium helper could be in ArtPiece

```ruby

class ArtPiece
   ...

  def self.find_by_artist(artist)
    ...
    art_pieces = get_json("/artists/#{artist.id}/art_pieces.json")

    ...
    (art_pieces['data'] || []).map do |ap|
        next unless ap
        data = process_json_data(ap)
        ArtPiece.new({ medium_id: data['medium']['id'] })
      }.compact
    end
     
    ...
  end
```

I just scribbled all this so it may not work but hopefully the idea comes across.

kneyzberg:
Closing Pull request as changes made in different branch

PR #45, opened 2023-10-06
Move JSON API unpacking logic into MauModel

bunnymatic:
problem
--------

The JSONApi format we fetch from MAU is kind of a mess.  But every model we build shouldn't have to know how to unpack it.  Some of the basics like 'attributes' and 'relationships' and 'id' should be unpackable without caring about the data type.

solution
--------

This PR shows a possible solution to let the MauModel own the responsibility of unpacking the JSON Api data without any knowledge of the consumer model.

It also updates the consumer model to pull data that *it* knows about (in this case `Artist`) from the base `id`, `attributes` and `relationships` (not used here but you could see it).


models/artist.rb
```diff
@@ -3,92 +3,79 @@
...
+  attr_reader :model, :id
 
-  def initialize(model_data)
-    @model = model_data
-  end
-
-  def id
-    @model['id']
+  def initialize(json_api_data)
+    puts 'J', json_api_data
```
bunnymatic, line 9: whoops - let over puts.  this should go away.

models/artist.rb
```diff
@@ -3,92 +3,79 @@
...
-      if fetched_artists.key? 'data'
-        clean_artists = fetched_artists['data'].map do |artist|
-          { 'id' => artist['id'].to_i, **artist['attributes'], 'artist_info' => artist['attributes']['artist_info']['artist_info'] }
-        end
-      end
-      SafeCache.set('artists', clean_artists)
-      artists = clean_artists
+    raw_artists = SafeCache.get('artists')
+    if !raw_artists || raw_artists.empty?
+      raw_artists = get_json("/artists.json?studio=#{studio.slug}")
```
bunnymatic, line 43: look at how clean this is ! :D

models/mau_model.rb
```diff
@@ -1,6 +1,24 @@
...
+  class BadFormat < Error; end
+  class JsonApiCollectionError < Error; end
+
+  attr_reader :attributes, :relationships, :id
+
+  def initialize(json_api_data)
+    data = json_api_data.with_indifferent_access
+    raise BadFormat, "This doesn't look like a JSON Api payload" unless data.key?(:data)
+
+    @id = data[:data][:id]
```
bunnymatic, line 14: here we just pull out the basics of the Json API data that we know about - independent of which model is consuming.

models/mau_model.rb
```diff
@@ -1,6 +1,24 @@
...
+
+    raise JsonApiCollectionError unless data.dig(:data, :attributes)
```
bunnymatic, line 16: Right here we *could* switch on something like `if data key exists, else if attributes key exists ...` which would handle the collection vs individual fetch.

bunnymatic (commented):
this is a super draft and does not yet handle "fetch all artists with 1 json call and then jam them in here" but it feels like it could without too much work.

models/artist.rb
bunnymatic: Also - this only updates `find` so a mentioned in the overall comment, `all` would need a little rejigger.

bunnymatic:
closing in favor of #51

PR #46, opened 2023-10-07
Move to Vite.js for frontend things

bunnymatic:
problem
-------

We have no good frontend story.  Definitely not one from this decade.

solution
--------

introduce [Vite.js](https://vitejs.dev).

This PR adds `vite_ruby` gem and with a little futzing and a small monkey patch gets it working in both admin and application side to serve up the css and javascript that we need.

It will certainly need some more love when we get to deployment but for a bit this should be a nice add.

To make things easier, it'll probably mean moving to using `overmind` or `foreman` to start everything up as we want the dev server and the app to be running together.

Note: this also adds a node/yarn requirement.

changes
---------

* Add vite_ruby
* move css and js into `frontend` directories and compile them with vite
* dump the old markit up and include with CDN links SimpleMDE editor which fits our pico styling way better.
* remove unused css
* update readme a bit


lib/vite_ruby/manifest.rb
```diff
@@ -0,0 +1,33 @@
...
+# Vite-ruby was serving up `/vite-dev/@fs/Users/jon/...` with a full path
+# but the dev server wasn't responding to this.
+# so... simplify this for now
+
+# rubocop:disable Style/ClassAndModuleChildren
+
+# IF we don't ignore this rubocop ClassAndModuleChildren rule,
+# this monkey patch doesn't work
+class ViteRuby::Manifest
+  def path_for(name, **options)
```
bunnymatic, line 13: not a *huge* fan of these monkey patches but it seems that either path helpers or something else are likely more sophisticated in padrino/hanami/rails and so we just need a little tweaking here.

I proved these out by running with dev server on and trying by manually building with RACK_ENV=production and then running without the dev server.  Both cases seem to work, and the feature tests pass on CI so I think it'll fly.

bunnymatic: interestingly, we had to patch `resolve_absolute_entry` because it's used in `lookup!` and there was some mismatch between the app root and the js.  I wonder if it has to do with the fact that we don't have a top level `app` dir here that is sitting next to `frontend` which makes these finders think they should step up one and then back down?

it was reporting "can't find `../vite/assets/application-SHA.js` in manifest which only had `/vite/assets/application-SHA.js`.  And it generates the `..` because it says something like `app.root_path.relative_path_from(file)` which is comparing the project `/` with `/frontend`.  If the app code was in `app/` maybe this would have just worked ?

PR #51, opened 2023-10-09
Kn/json api helper update

kneyzberg:
### Problem 
- currently, our ArtPiece model was not accessing the medium Id related to Art Pieces.
- the data processing of JSON data in Artist and ArtPiece model was too verbose and clunky.
- Initial proposal of a JsonApIHelper was not generalized enough

### Solution
-  MauModel process and unpack all json data into id, relationships, and attributes. 
-  Artist, Studio, and ArtPiece models inherit from MauModel and access data through att_readers, making code cleaner and more readable
- Removed unnecessary methods from Models.


models/studio.rb
```diff
@@ -2,35 +2,29 @@
...
-    model['slug']
+    attributes['slug']
   end
 
   def artists
-    model['artists']
+    relationships['artists']['meta']['count']
   end
 
   def to_param
```
bunnymatic, line 27: I wonder if we use the thing anywhere?  I'm ok with leaving it but we could likely trim this even further.  

not a blocker just a thought.

bunnymatic (approved):
Looks good.

I like that you didn't change any tests and made this pretty good sized refactor.

models/studio.rb
kneyzberg: i don't think we use it anywhere at the moment! Happy to merge this and then trim it down even more

PR #61, opened 2023-10-17
updated Artist page css to match current 1890 bryant

kneyzberg:
Screen Shot 2023-10-16 at 11 55 25 PM
### Problem 

Currently our layout on the page is not responsive and outdated. We need to get something that looks visually more like the current 1890 bryant street studios website.

### Solution 
- moved Artist page css into it's own file.
- removed breadcrumbs and contact pages from slim template.
- added methods on ArtPiece and Artist to get pertinent information.
- updated css to look more like current 1890 bryant.


frontend/stylesheets/artist.scss
```diff
@@ -0,0 +1,38 @@
...
+
+  .artist-page {
+    display: flex;
+    flex-direction: row;
+    gap: 50px;
+  
+
+    section.info, section.bio {
+      flex: 1;
+      font-family: Alata;
```
bunnymatic, line 12: This doesn't actually work unless we include the font in our assets.  You can see the difference here (us on the left, 1890 on the right).  Let's leave the font definition out.  I think the default from `pico` which is san-serif something looks good enough.



Screen Shot 2023-10-17 at 8 01 40 AM

bunnymatic (commented):
I like that it's responsive but weirdly on the phone it doesn't become single column.  kind of weird.  We can file a separate ticket to figure that out.

Screen Shot 2023-10-17 at 8 07 43 AM

views/artist.slim
```diff
@@ -1,28 +1,18 @@
...
-            = [ap.medium, ap.year].compact.join(", ")
           .piece
-            img src=ap.images['medium']
+            img src=ap.images['large']
+          .title = ap.title
+          .medium_year
+            = [ap.year, ap.medium, ap.dimensions, ap.price].compact.join(" - ")
     section.bio
       h1 = @artist.fullname
+      h3 = 'Studio ' + @artist.studio_number
```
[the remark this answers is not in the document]
kneyzberg: yup, makes sense, will do!

frontend/stylesheets/artist.scss
kneyzberg: Copy that, will do!

bunnymatic:
LGTM

bunnymatic (approved):

PR #62, opened 2023-10-19
mobile web ready

bunnymatic:
problem
------

the mobile web layout shows 2 columns but we know the phone is a small
browser.

we should show 1 column.

solution
-----

add the right meta tag to the header so mobile browsers basically know
their size/shape.

image

Also - dump `home` link in nav and let a click on the logo to take us "home"


kneyzberg (approved):
Awesome, makes sense

PR #63, opened 2023-10-21
Limit number of art pieces for each artist

bunnymatic:
problem
---------

we know the MAU api is a little slow to get art piece images.
so let's not fetch all of them.

This corresponds with https://github.com/bunnymatic/mau/pull/388
on Mission Artists where we add the ability to limit this call.

solution
---------

* update the `ArtPiece.find_by_artist` to `ArtPiece.all_by_artist`
* add `count=3` to that call so we only get 3 art pieces
* update cassettes - some re-recording and some by manually updating the
  cassette to match the new call endpoint pattern


spec/app_spec.rb
```diff
@@ -62,7 +62,7 @@ def app
     it 'shows the artist\'s name in the title' do
       visit '/artists/1'
       name = find('.bio h1')
-      expect(name.text).to match 'Huels LLC'
+      expect(name.text).to match 'Eelva Blake'
```
bunnymatic, line 65: here we re-recorded so new first name on this one test.

PR #66, opened 2023-10-24
Bundle update

bunnymatic:
Run bundle update.

updates active record, rubocop and a couple other things

after this goes live, you'll want to 'bundle update'


kneyzberg (approved):
Looks good to me

PR #67, opened 2023-10-30
Kn/update contact page

kneyzberg:
### Problem 
Contact page looks outdated and needs to match more to the current version of 1890 Bryant .

### Solution
- updated icon pallet to include facebook and instagram icons.
- renamed contact page to `connect.slim`.
- updated contact page with new icons.
- updated contact page with embedded google map of 1890 address.
- added constants for instagram and fb url.


views/connect.slim
```diff
@@ -0,0 +1,12 @@
+.content
+  == render_slim_partial :breadcrumb, locals: { breadcrumbs: [ "contact" ] }
+  h1 Contact Us 
+
+  div.connect-icons
+    a href="#{INSTAGRAM_URL}" style="text-decoration:none" aria-label="Instagram Icon"
```
bunnymatic, line 6:
```suggestion
    a href=INSTAGRAM_URL style="text-decoration:none" aria-label="Instagram" 
```
We're in slim which is actually ruby so we don't need to quote *and* interpolate with `#{}`.

The label should be like the replacement for text that might be there for the icon so we don't need th word "icon".  This is just a link to "instagram".

And let's put the `style` in the css instead of inline.

something maybe like
```
.connect-icons a {
   text-decoration: none;
}
``` 
in some appropriate css file.

views/connect.slim
```diff
@@ -0,0 +1,12 @@
+.content
+  == render_slim_partial :breadcrumb, locals: { breadcrumbs: [ "contact" ] }
```
bunnymatic, line 2: let's dump the breadcrumb and h1 here.  we'll need to update the test to look for something else like maybe find the link to instagram on the page (instead of `have_content 'Contact Us'`

frontend/stylesheets/connect.scss
```diff
@@ -0,0 +1,14 @@
+.connect-icons {
+  font-size: 6rem;
+  display: flex;
+  flex-direction: row;
+  justify-content: space-evenly;
+  padding-bottom: 20px;
+}
+
+
+@media (max-width: 800px) {
```
bunnymatic, line 10: we probably could leave this.  These icons are small enough that they fit ok as a row even as the screen gets smaller.

views/connect.slim
kneyzberg: Understood, will look for the href a tag

frontend/stylesheets/connect.scss
kneyzberg: Making it smaller seems to look better, i'll do that.

bunnymatic (approved):

PR #68, opened 2023-10-31
Updated main layout to container-fluid

kneyzberg:
### Problem 
Content on our pages, especially on artist page had big gaps and was not filling up the screen correctly. 

### Solution
Update CSS of main layout to use container-fluid to allow for more fill of the screen. 

- updated .main tag to use container fluid instead of container
- created gutters to allow for margins are pages
- updated artist page css to have better margins and look more centered
- updated home page image padding so it takes up full page


frontend/stylesheets/main.scss
```diff
@@ -93,3 +93,15 @@ main {
     flex: 1;
   }
 }
+
+.gutters {
```
bunnymatic, line 97: looks good.  i wonder... did we put `gutters` on everything?  if so, should this just be attached to `.content` or `main.container`?  maybe we need it to be attachable so if so, this is fine.  just asking

bunnymatic (approved):

frontend/stylesheets/main.scss
kneyzberg: No it totally should be a part of content, it makes more sense that way. For some reason was under the impression that .content was a pico class and didn't want to overwrite anything, but I see where content is now, will adjust

PR #70, opened 2023-11-04
Support fake cache in tests

bunnymatic:
Problem
----------

we thought we were using a fake cache but it wasn't *actually* working
during test mode.

solution
---------

build a better cache init system that for an environment allows us to
specify the cache type.

This assumes a local memcache at the standard memcache port.  when we
get to deployment, we may need a little more config stuff but for now
this seems to do the trick

testing?
--------------------

* ran this and shut off memcache locally - works fine - it just means it
keeps calling mau
* turn memcache back on and everything starts getting cached
* run tests and prove that they work

Inadvertently, it seems that it's reduced the logging output for tests.


lib/safe_cache.rb
```diff
@@ -26,33 +28,21 @@ def self.flush
...
-                  ENV['MEMCACHIER_SERVERS'].split(','),
-                  {
-                    username: ENV.fetch('MEMCACHIER_USERNAME', nil),
-
-                    password: ENV.fetch('MEMCACHIER_PASSWORD', nil)
-                  }
-                )
-              else
+    @client = case cache_client
+              when :dalli
```
bunnymatic, line 42: this will need updating when we get to a deployment but for now... no problem

bunnymatic:
close in favor of #98 
this pr had too much crap in it as it was.

PR #80, opened 2023-11-09
First Pass art_piece_card render and styling

kneyzberg:
### Problem 
Currently, our Artist Page does not have an Artwork section that shows the Artist's other art pieces like current 1890 Bryant. 

### Solution
- created artpiece_card element and rendered the partial on the artist page for each piece of artwork. 
- created Artist Artwork Section below the Artist Details on Artist page. 
- created artpiece-grid layout and added styling for the grid and grid items.

Currently the `<a> tags` direct back to the artist page, this will change once we have the artpiece pages set up.


views/artist.slim
```diff
@@ -15,3 +15,10 @@
         h3 = 'Studio ' + @artist.studio_number
       == markdown_content(@artist.bio)
       == render_slim_partial "artists/social", locals: { artist: @artist }
+  section.artist-artwork
+    h2 Artwork
+    hr
+    ul.artpieces-grid
+      - @artist.art_pieces[1..].each do |ap|
```
bunnymatic, line 22: let's just include all so it acts like a carousel sort of.
it almost feels like the artist/art_piece show could be the same page and we just say which art piece to show.  we can discus but for now, since we know we're limited to ~3~ 4, let's just show all of them in the thumbs.
```
@artist.art_pieces.each...
```

alternatively, if we leave it at 3, then we know that set will never be more than one row on a big screen.

alternatively, maybe we limit to 5 and let it really grow... it's a battle of performance v visuals.

models/art_piece.rb
```diff
@@ -40,7 +40,7 @@ def thumbnail
     images['small']
   end
 
-  MAX_NUM_ART_PIECES = 3
+  MAX_NUM_ART_PIECES = 4
```
bunnymatic, line 43: +1

frontend/stylesheets/artist.scss
```diff
@@ -27,5 +27,39 @@
...
+    border-top: 1px dotted grey;
+    margin-bottom: 50px;
+  }
+
+  .artpieces-grid {
+    flex: 1;
+    display: grid;
+    grid-template-columns: repeat(1, 1fr);
+    gap: 20px;
+    @media screen and (min-width: layout.$screen-sm-min) {
```
bunnymatic, line 50: we might want to go down to 1 at small screens.  here's a screenshot that I see at 574px wide

image

might be better as one.  also maybe our break points aren't great.  but worth a check.

not a blocker, we can tweak this stuff if this continues to blossom.

bunnymatic (approved):
I think it looks great.   I'd just merge it and we can tweak with future pr's.

bunnymatic:
oh... yeah - we can do it in a follow up but we need some tests.

```
visit '/artist/<id>'
expect(page).to have_css('.artist-card')
expect(page).to have_css(art_piece.title)
```
or something in that realm.

bunnymatic:
Looking at the current test failures, it appears you'll have to prove that out with tests already.  so maybe it'll be with this pr ;)

views/artist.slim
kneyzberg: Makes, will leave at 4 for now and add all of them back in, and we can discuss.

frontend/stylesheets/artist.scss
kneyzberg: Yup, it looks way better! I added 2 columns for medium screens which I think looks better.

models/art_piece.rb
```diff
@@ -40,10 +40,10 @@ def thumbnail
...
 
   def self.all_by_artist(artist)
-    art_pieces = Mau::JsonApiClient.new.get_json("/artists/#{artist.id}/art_pieces.json?count=#{MAX_NUM_ART_PIECES}")
+    art_pieces = Mau::JsonApiClient.new.get_json("artists/#{artist.id}/art_pieces.json?count=#{MAX_NUM_ART_PIECES}")
```
bunnymatic, line 46: This feels a little suspicious to me.  I have a feeling that this change will break a bunch of the VCR tests because the path will look a little different than what we recorded before.  Did you do this to get rid of a `//` in the path?  or did you do this to get tests working?

I think you might want to revert this change and the one below where you took out the leading `/`, revert all your fixture changes and then re-record from that point.  You might get closer to the end.

What I think might be happening is that you made this change, re-recorded some of the fixtures so your new tests with the new recordings work, but the old recordings are also using this code , so those are saying "i can't find a matching request url" because you took a `/` out of it, and so they are failing.

just a guess

spec/app_spec.rb
```diff
@@ -81,7 +81,19 @@ def app
       expect(page).to have_link('', href: 'https://www.instagram.com/explore/tags/linocuts') do |link|
         expect(link['aria-label']).to eq 'Instagram'
       end
-      expect(page).to have_link('http://gottliebhaley.org/rosette.predovic')
+      expect(page).to have_link('http://doyle.example/landon')
+    end
+
+    it 'shows the Artwork Section' do
+      visit 'artists/1'
```
bunnymatic, line 88: Similarly you have this `artists/1` and below `/artists/1`.  These should be consistent. Looking above it looks like we favor with the leading slash.

PR #81, opened 2023-11-11
Find and use a JSONAPI deserializer

bunnymatic:
problem
--------

we have 2 issues.

1. low level tests driven by vcr are fussy
1. `jsonapi-rb` which is the library MAU is using doesn't have a good
   deserializer so we had to write a bunch of code to do that ourselves.

solution
---------

This PR takes 2 steps to
1. install a decent JSONAPI library whose deserializer actually works -
   which means we can remove a bunch of our model initializer code that
   initially went in with #12, #38, #44 and maybe a few other prs.
2. separate the model code from the fetching code so we can be more
   strategic about when to use VCR
3. build some factories that generate JSON API structures so we can test
   stuff without VCR

Changes
----------

### Separate repos for fetching from model code

This work means our models are just data thingies and the consumers that
want to fetch records need to go through one of these Repos.  It also
means we can move our model tests to simply generate json data, build
the model, and check the attributes.  This removes the need for VCR in
the model tests.

Then we can test the fetching logic in the Repo tests.

### Build factories for more flexible testing

Though a little painful, I was able to build factories that allow us to
construct without too much trouble, artists, and art pieces without
relying on VCR.  Instead the factories will generate a jsonapi payload
that looks right for the different model and then we can just construct
new instances of the model that way

It only removes the VCR from the model tests but I will look (in future
PR's) to see if there is away to push that further to remove more VCR
dependencies.

### Use JSONAPI deserializer

This simply removes all the craziness we did around relationships and
attributes and whatever and leaves it to a library that was built to do
that work.

Testing
---------

Tested locally with memcache off and a bunch of testing IRL.


models/artist.rb
```diff
@@ -5,75 +5,70 @@
...
 
   def art_pieces
-    attributes['art_pieces'][0..3].filter_map do |item|
-      ArtPiece.find(self, item['id'])
+    @model['art_pieces'][0..3].filter_map do |item|
+      Repos::ArtPieceRepo.find(self, item['id'])
     end
   end
-
-  def self.find(id)
```
bunnymatic, line 69: moved to `Repos::ArtistRepo`

models/mau_model.rb
```diff
@@ -1,27 +1,15 @@
...
+require 'jsonapi'
 
-  attr_reader :attributes, :relationships, :id
+class MauModel
+  include JSONAPI::Deserialization
 
   def initialize(json_api_data)
-    data = (json_api_data || {}).with_indifferent_access
-
-    raise BadFormat, "This doesn't look like a JSON Api payload" unless data.key?(:data) || data.key?(:attributes)
```
bunnymatic, line 13: look at all this we don't have to do!!!

views/artist.slim
```diff
@@ -8,9 +8,9 @@
             img src=ap.images['large']
           .title = ap.title
           .medium_year
-            = [ap.year, ap.medium, ap.dimensions, ap.price].compact.join(" - ")
+            = [ap.year, ap.medium, ap.dimensions, ap.price].reject{|v| v.nil? || v.empty?}.join(" - ")
```
bunnymatic, line 11: minor nit - this was showing `2002 - ` if you had an empty string in one of these so... a little smarter compact function.

kneyzberg: ahh that makes sense, thanks!

kneyzberg (commented):
Wow this is way better I think for the specs and makes so much more sense! Our tests will now be in sync!

PR #86, opened 2023-11-11
Add ability to fill empty images with placeholders

bunnymatic:
problem
----------

a bunch of test data from mau ends up with empty images

solution
----------

let's just add placeholders on dev if the images are empty so at least
we can see what the layout looks like.

I mean look how cute this is

image


models/artist.rb
```diff
@@ -67,7 +67,7 @@ def instagram
   # end
 
   def art_pieces
-    @model['art_pieces'][0..3].filter_map do |item|
+    @art_pieces ||= @model['art_pieces'][0..3].filter_map do |item|
```
bunnymatic, line 70: I found that this pr busted the tests because it introduced 2 more calls to MAU (with the cache off) but they are the same.  So if we locally just memoize in the model, we save a few actual calls.

which also explains why we have some fixture files bloated with `artists/1.json?count=4` over and over and over

views/artist.slim
```diff
@@ -15,10 +15,11 @@
...
       == markdown_content(@artist.bio)
       == render_slim_partial "artists/social", locals: { artist: @artist }
-  section.artist-artwork
-    h2 Artwork
-    hr
-    ul.artpieces-grid
-      - @artist.art_pieces[1..].each do |ap|
-        li.artpiece-grid--item
-          == render_slim_partial :artpiece_card, locals: { art_piece: ap, artist: @artist }
+  - if @artist.art_pieces.length > 1
```
bunnymatic, line 18: don't show artwork section unless there are 2 or more pieces

PR #91, opened 2023-11-12
Force tests to run in test mode

bunnymatic:
problem
---------

our entire environment runs in `development` mode all the time.
This means the work to setup a fake cache was a flop.

solution
---------

Let's do our best to force running tests in `test` mode

This pr forcibly sets `ENV['RACK_ENV']` in `spec_helper` before all
requires so that we hopefully get that out of the way and everyone
(including the app) follow suite and pick up this change.

You can still override with `RACK_ENV=whatever_crazy_environment
bin/rspec` but we'll warn you that it's probably not a good idea.


config.ru
```diff
@@ -19,11 +19,4 @@ puts "Proxy? #{ViteRuby.run_proxy?}"
 # setup static serving
 use Rack::Static, urls: ['/images', '/stylesheets', '/javascripts'], root: File.join(root, 'public')
 
-# Paperclip.configure do |config|
-#   config.root               = root # the application root to anchor relative urls (defaults to Dir.pwd)
```
bunnymatic, line 23: just random clean up

PR #94, opened 2023-11-12
Add Navigation structure

bunnymatic:
problem
------

As we get to adding another navigation element, I wanted to refactor the
way we handle navigation to be a little more OO and easier to use.

solution
----------

build a base `Navigation` class which holds a list of navigation
`Item`s.  This object is iterable so we can just run through the nav
elements and render the links.

Also provide (on the Item) a helper to ask "are you on this page?"  Is
it `current?`

changes
-------

* add services directory to distinguish between library stuff and
  classes that are specific to this application
* add `Navigation::Base`
* add `Navigation::Item`
* build `AppNavigation` and `AdminNavigation` specific instances
* use those in rendering nav for both admin and app nav bars.
* rename `admin/_admin_nav` to `admin/_navigation` to match app level file `_navigation.slim`


lib/navigation.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module Navigation
+  Item = Struct.new(:path, :display_name, :key, :current, keyword_init: true) do
+    def initialize(path:, display_name:, key:, current: false)
```
bunnymatic, line 5: if we start getting mobile and we need icons or something we can add to this struct.

PR #96, opened 2023-11-13
Explore Artist Link and Artist Shuffle

kneyzberg:
### Problem 
- We want to have a way to surface artists randomly, so that the Artist Page can link to a random other Artists page, and be able to navigate to that on the Artist Page.
- Artist order on the `/artists ` page should be random and shuffle on every visit. 

### Solution
- changed prev/next links to Explore Artists anchor tag
- added shuffle method to artist_repo to shuffle artists
- updated /artists/:id page to generate a random artist to link to
- updated /artists page to generate random shuffle order on every visit
- added specs for Artist_Repo.shuffle method
- added specs to app_spec to account for new shuffle method in view specs
- updated fixtures


frontend/stylesheets/artist.scss
```diff
@@ -3,6 +3,38 @@
...
+  a {
+    text-decoration: none;
+  }
+ 
+  .artist-nav--link {
+    display: flex;
+    align-items: center;
+  }
+
+  .artist-nav--prev--icon {
```
bunnymatic, line 19: should we dump these `--prev` and `--next` and `--slash` definitions?

if we kept them, I might argue for the upper component to control things like
```
.artist-nav--text {
   > span {
      margin-right: 5px;
   }
   > span:last { 
     margin-right: 0;
   }
```

this syntax might not be totally right but we shouldn't need to define spacing of all 3 elements in the thing.
also - in this case we could maybe just put a space " " between things in the slim to let it just space it self out.

so many words.  in the end, if we don't apply the class, let's dump the rules so no one is confused.

spec/app_spec.rb
```diff
@@ -64,17 +64,32 @@ def app
       expect(page).to have_css('.artists-grid--item .artist-card--image')
       expect(page).to have_css('.artists-grid--item .artist-card--details')
     end
+
+    xit 'shuffles the artists so there is a different order' do
```
bunnymatic, line 68: can we un `xit` this test?  I like what it's trying to test.  not a blocker.  we can leave it if there's an effort to actually prove this out.

frontend/stylesheets/artist.scss
kneyzberg: for sure, makes sense to dump until we bring back the Prev/Next

spec/app_spec.rb
kneyzberg: this has been fixed now that i have more data to work with, the change of them being the same is slim :)

PR #97, opened 2023-11-13
Even better (sort of) logging

bunnymatic:
problem
------

with #95 we thought we'd solved the issue.  everything got logged!

but it turns out, that took it *out* of the stdout/err during the dev
server.

solution
------

Update logging so that in development errors are in the console, but for
other environments (test, prod) they end up in the log files.


bunnymatic:
Reviewers (@kneyzberg) 
maybe you should pull this branch and try it out.
put some errors in the flow and try dev/test and see if it acts the way that makes sense to you - before we merge.

kneyzberg:
Finally taking a look at these! Logging for dev is awesome, so so much better! And test seems fine as well! Thank you, this feels good to merge to me

kneyzberg (approved):

PR #106, opened 2023-11-20
Kn/restore prev next links

kneyzberg:
### Problem 
- needed to restore prev/next links on artist page and have them point to artist pages in a stable order of artists being sorted alphabetically by last name
- need to remove Explore Artist Link

### Solution 
- removed explore artist link
- added prev/next links to artist.slim and updated the css
- created a .find_neighbors method in Artist_repo to do business logic of sorted artist by last name and finding prev/next artists in the list
- updated specs to test .find_neighbors and updated view test for artist page
- removed breadcrumb from /artists page


repos/artist_repo.rb
```diff
@@ -21,6 +21,18 @@ def self.shuffle(excluded_artist_ids: [])
...
 
+    def self.find_neighbors(id)
+      artist_ids = all.sort_by(&:last_name).map(&:id)
+      artist_idx = artist_ids.find_index(id)
+      prev_idx = artist_idx - 1
+      next_idx = artist_idx + 1
+
+      prev_id = prev_idx >= 0 ? artist_ids[prev_idx] : artist_ids[artist_ids.length - 1]
+      next_id = next_idx < artist_ids.length ? artist_ids[next_idx] : artist_ids[0]
+
```
bunnymatic, line 32: This doesn't need to be changed but I wonder about...
```
      prev_idx = artist_idx - 1
      next_idx = artist_idx + 1
      if prev_idx < 0
          prev_idx = artist_ids.length - 1
      end
      if next_idx > artists_ids.length
          next_idx = 0
      end

      {prev: artist_idx[prev_idx], next: artist_id[next_idx]}

```

this is a minor refactor but I think it might be a little easier to read/reason about.  I took me a minute looking at the ternary to say "are these right?  should it be <=?".  They *are* right but it was a little tough to see.  

So... up to you.  maybe something to think about.  or we can update it later too.

bunnymatic (approved):

repos/artist_repo.rb
```diff
@@ -21,6 +21,18 @@ def self.shuffle(excluded_artist_ids: [])
       all.reject { |a| excluded_artist_ids.include?(a.id) }.shuffle
     end
 
+    def self.find_neighbors(id)
```
bunnymatic, line 24: forgot to say - i like this method.

repos/artist_repo.rb
kneyzberg: No you're right this is wayyy clearer and easier to read, will fix and implement it this way.

PR #107, opened 2023-11-20
Kn/change artpiece on show page

kneyzberg:
### Problem 
Currently we have no way of seeing a bigger version of the thumnail images on the Artists show pages. We want to see the thumb image show up as the main image when we click on it. 

### Solution 
- added logic for parsing query parameters that include art_piece_id on `aritsts/:id` show page. 
- added logic that chooses  a `main_art_piece` based on if query param exists or not.
- changed logic to ensure Artist Artwork section does not include main art piece.
- added view specs to test functionality of links with query params.


views/artist.slim
```diff
@@ -27,6 +27,6 @@
       h2 Artwork
       hr
       ul.artpieces-grid
-        - @artist.art_pieces[1..].each do |ap|
+        - @artist.art_pieces.reject{|ap| ap.id == @main_art_piece.id }.each do |ap|
```
bunnymatic, line 30: You could put this in the "controller" method too as `@secondary_art_pieces = `...

i also wonder if we should just show everything below.  Maybe not.  it can be a separate discussion.

bunnymatic (approved):
so easy!

views/artist.slim
kneyzberg: Yeah I was thinking the same thing actually, will fix!

PR #112, opened 2023-11-28
Separated routes and specs into separate files

kneyzberg:
### Problem 
As our application gets bigger and we add more routes, we have a need to separate routes into separate files more clear separation of concerns, organization, and to reduce the risk of merge conflicts. 

### Solution
- created `controllers` and `controllers/admin` directories and moved all routes into separate controller files
- created a `spec/controllers` directory and moved all specs in admin_spec.rb and app_spec.rb into separate files mirroring the controllers directory.


controllers/admin/artist_exc_controller.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+module BryantStreetStudios
+  class App < Sinatra::Base
```
bunnymatic, line 4: No need to shorten this - let's be clear and use long clear file names.

`artist_exclusions_controller.rb`

controllers/admin/artist_exc_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+    ### Admin artist exclusions controller
+
+    ### artist_exclusions
```
bunnymatic, line 7: we could probably dump these comments now that we have only a few routes, or maybe rename the to match the "view" so this would be

```
### index
```
and below
```
### create
```
``` 
### delete
```

controllers/admin/image_resources.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
```
bunnymatic, line 1: rename `image_resources_controller.rb`

controllers/admin/markdown_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+
+    ### preview markdown do
+    post '/admin/markdown' do
+      require_authorization!
+      markdown_content(params[:data])
+    end
+
+    # Admin cache Controller
+
+    get '/admin/cacheflush' do
```
bunnymatic, line 15: this should be in a different place - 
I might think
```
admin/cache_controller.rb
```
and then maybe we update the path to 
```
get '/admin/cache/flush' do
```
which would allow us to add new cache methods like `warm` which might be handy down the road.

db/schema.rb
```diff
@@ -12,38 +10,39 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.1].define(version: 20_231_125_022_114) do
+ActiveRecord::Schema[7.1].define(version: 2023_11_25_022114) do
   # These are extensions that must be enabled in order to support this database
-  enable_extension 'plpgsql'
+  enable_extension "plpgsql"
```
bunnymatic, line 15: we should probably align and decide on this.  I think here the only difference is that I rubocop'd my schema file and you migrated and did not.

I'm fine with either one but to avoid churn on this file...

for this PR, you can either
1. `bin/rubocop -A db/schema.rb` and check it in and see if it becomes a no-op (no difference)
2. leave it here in this PR and we can align and maybe add rules to ignore it in rubocop or something so we don't get the churn in future prs.

spec/controllers/admin/artist_exc_controller_spec.rb
```diff
@@ -0,0 +1,75 @@
+# frozen_string_literal: true
```
bunnymatic, line 1: rename `artist_exclusions_controller_spec.rb`

bunnymatic (commented):
Generally looks really good.  seems like a pretty simple move.  I like the busted up specs.
Should/could the `app_spec` also be broken into bits?  You might be able to copy and paste fixtures into the right directories if you wanted to avoid re-running vcr.  it's a little tedious, but something that could happen.


I left a couple notes.  

1. a couple of file renames for consistency
2.  I think if we want to leave the method comments like
```
# index
get '/artists' do

...
# show
get '/artists/:id' do
```

we should touch every controller and make them consistent.
same `#` or `###` lead comment, and consistent language which I think should be one of the following: index, show, new, create, update, delete or a speciality one like flush for the cache controller .

Alternatively, because the files are smaller and you can read the whole thing, just dump all the identifier comments in all controllers.

controllers/admin/artist_exc_controller.rb
kneyzberg: For consistency, i'll add them to match the views

controllers/admin/markdown_controller.rb
kneyzberg: Yup, this should totally be separate, missed this one. Sounds good

db/schema.rb
kneyzberg: Will do option 1, copy that

kneyzberg:
@bunnymatic  For the `app_spec` do you mean the `admin_routes_spec.rb` file? Meaning we would copy that test into every single controller? I'm not against that idea, i guess it would be cleaner/more separated.

controllers/admin/content_blocks_controller.rb
```diff
@@ -0,0 +1,52 @@
+# frozen_string_literal: true
+
+module BryantStreetStudios
+  class App < Sinatra::Base
+    ## Admin content blocks controller
+
+    ### show
```
bunnymatic, line 7: strictly speaking this is `### index`

controllers/admin/hero_images_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+    ### index
+    get '/admin/hero_images' do
+      @current_section = 'hero_images'
+      require_authorization!
+      @current_hero = Repos::HeroImageRepo.current
+      @hero_images = Repos::HeroImageRepo.all
+      render_slim_admin_page 'hero_images/index'
+    end
+
+    ### delete
```
bunnymatic, line 16:
```suggestion
    ### activate
```

controllers/admin/hero_images_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+    get '/admin/hero_images/:image_resource_id/activate' do
+      require_authorization!
+      Repos::HeroImageRepo.activate!(params[:image_resource_id])
+      redirect '/admin/hero_images'
+    end
+
+    ### delete
```
bunnymatic, line 23:
```suggestion
    ### deactivate
```

controllers/admin/image_resources_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+      require_authorization!
+
+      tempfile = params[:file][:tempfile]
+      filename = params[:file][:filename]
+      Repos::ImageResourceRepo.save(filename, tempfile)
+      redirect '/admin/image_resources'
+    end
+
+    ### delete
+    get '/admin/image_resources/:slug/delete' do
```
bunnymatic, line 34: in a later PR maybe we can make these actually be `delete` http method instead of `get`.  not sure how to do that in Sinatra and we'll need to update the javascript cuz it can't be just a link but it'd be more HTTP accurate.

not for now.  just kind of note to ourselves.

bunnymatic (approved):
Nice.  

Approved with notes that are super minor and could be ignored or fixed up and merged without another review.

Re: the `app_spec`, I didn't mean the routes.  I like the way you broke out the `admin_routes` shotgun spec for authorization.  We could argue that everything you moved into controller specs is actually more like a feature spec because we are using `visit` which is hitting things through a browser. 

I think I wondered if anything in the `app_spec` is just a controller spec for "artist controller" (or other non admin controllers).  I suppose where we use `visit` it's really more than just a controller spec so it could stay and anything that does something like `get '/artists'` that's more in line with a controller spec.

for now, just leave it.  we can talk about controller specs vs "feature" or "e2e" specs and maybe decide if we re-org.  for now I'd just leave it.  things are being tested at least as well as they were before so ... that's what we want.

controllers/admin/image_resources_controller.rb
kneyzberg: Copy that, will make note of that.

PR #113, opened 2023-11-29
Fix issue with images that have single quotes

bunnymatic:
problem
------

if you pull data from MAU prod, you'll find that the artist "Eliot
Rattle" has an art piece with a filename "That's_all_folks.jpg".  MAU
doesn't seem to correctly escape or replace single quotes in filenames.

When we try to jam this in a `background-image: url()` tag, the single
quote in the HTML throws things off.

solution
------

This pr replaces fetching images via the `images` hash and instead makes
an `image` method which takes `size` and returns the url that has single
and double quotes escaped.

This works for `Eliot Rattle`.  Hopefully, it'll work for any other
weird filename we might get from MAU.


views/artist.slim
```diff
@@ -5,14 +5,14 @@
       .artist-nav--text.prev Previous
     a.artist-nav--link href="/artists/#{@next_artist}" title="" aria-label=""
       .artist-nav--text.next  Next
-      .artist-nav--icon >    
+      .artist-nav--icon >
```
bunnymatic, line 8: my editor cleaning trailing whitespace.

models/art_piece.rb
```diff
@@ -27,6 +27,17 @@ def price
     @model['display_price']
   end
 
+  def image(size = :medium)
+    # escape single and double quotes
+    images[size.to_s].gsub("'", '%27').gsub('"', '%22')
```
bunnymatic, line 32: this is very close to a uri encode but that seems maybe like overkill so...

kneyzberg (approved):
Looks good to me

PR #114, opened 2023-11-30
update icons and use them in the `hero image` page

bunnymatic:
problem
-----

we already wanted a few more icons.

solution
------

so i pulled some in from the Material UI with ico moon and
set and then used the `flash` for the hero image activate icon.

changes
------

* add new font files
* update palette
* update hero image index to use the new icons
* update dev cache to a way longer expiry time
* fix carousel that was showing js lint issues.
* replace prev/next with icons


views/artist.slim
```diff
@@ -1,11 +1,11 @@
 .content
   .artist-nav
-    a.artist-nav--link href="/artists/#{@prev_artist}" title="" aria-label=""
-      .artist-nav--icon <
-      .artist-nav--text.prev Previous
-    a.artist-nav--link href="/artists/#{@next_artist}" title="" aria-label=""
-      .artist-nav--text.next  Next
-      .artist-nav--icon >
+    a.artist-nav--link href="/artists/#{@prev_artist}" title="Previous" aria-label="Previous"
```
bunnymatic, line 3: We should try to use sensible link titles/aria-labels for accessibility.  Especially when we're only using icons - as here we've moved from "< Previous" to "<"

PR #121, opened 2023-12-04
Added 1890 Instagram Profile Embed

kneyzberg:
### Problem 
We wanted to be able show a preview of the 1890 Instagram profile. 

###
- copied code from Instagram Embed feature on 1890 Bryant IG profile. 
- converted instagram embed code of 1890 Bryant to slim.
- created partial slim template for instagram embed code.
- rendered on home index page.


bunnymatic:
nice - couple questions?
1. can we make it fill the width of the page.
2. can we maybe wrap it with a config setting kind of like a feature flag.  I think I'd want to show Trish and decide if we want to show it.  it is super simple but now looking at it, the extra header is kind of annoying (compared with the 1890 Wordpress version). 
3.  Do we have any styling control?


Also given our other slack discussion, maybe you're other approach would solve these issues and might be worth pursuing... :D

bunnymatic:
I guess I think we could merge *if*
1. we can easily turn it off with a config flag or something that doesn't require a code push
2. we can make it fill the page width

bunnymatic:
don't worry too much about tests.  it's an embed.  we shouldn't test instagram.  but if it's tests like "when the flag is on, I see it', that's nice and simple and maybe worth doing.

bunnymatic (approved):

bunnymatic:
Also it might be interesting to have a test that overrides the config since I don't know if we do that.  Just as another sample of doing that.

PR #126, opened 2023-12-06
Move controller specs to spec/features

bunnymatic:
problem
-----

we don't actually do controller tests.  our tests are fully e2e
integration specs.  so let's call them as such.

solution
-------

to achieve this, we can
1. move admin spec heloers into a module and include them just for
   feature specs.
1. set the `:type` for feature specs
1. move all the specs from `controllers` dir to `features`

Note:
[rubocop-rspec](https://github.com/rubocop/rubocop-rspec) assumes Rails
even though we don't have it.  So we've disabled that rule - and also
filed an [issue](https://github.com/rubocop/rubocop-rspec) to see if
they might consider not including it if Rails is not in the project.


spec/features/connect_pages_spec.rb
```diff
@@ -2,23 +2,7 @@
 
 require_relative '../spec_helper'
 
-Dir[File.join(File.dirname(__FILE__), '..', '{lib,models}/**/*.rb')].each do |file|
-  require file
-end
-
 describe BryantStreetStudios::App, vcr: { record: :none } do
-  def login_as_admin
```
bunnymatic, line 10: If we looked closely, these tests don't need admin anything so this method is kind of just a waste of space.

spec/support/admin_feature.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+module AdminSpecHelpers
+  def app
```
bunnymatic, line 4: it's pretty easy to wrap up some Spec helpers that are only available given certain metadata or even just as helpers we want for rspec using this module technique in the the `spec/support` dir.  so to avoid all the repetition we had...

PR #130, opened 2023-12-09
Build more responsive artist page

bunnymatic:
problem
------

we had trouble with some data from MAU that blew out the right hand side
of the artist page - where the bio goes.

solution
------

As a first cut, force the max width of the 2 sections on the artist page
to be no larger than 42% of the page - this accounts roughly for gutters
and spacing and makes it so everything fits in the page.

Once i was in here, i looked into some responsiveness for the artwork
and the overall layout and revamped slightly the classes and css
settings to improve the visuals on the artwork at the bottom and make
sure that nothing falls off the page at different browser sizes.

screenshots
--------------

Screen Shot 2023-12-08 at 6 47 21 PM
Screen Shot 2023-12-08 at 6 47 33 PM
Screen Shot 2023-12-08 at 6 47 44 PM
Screen Shot 2023-12-08 at 6 47 51 PM


kneyzberg:
Looks good to me!

PR #134, opened 2023-12-09
Add some styling and sold tag to main art piece

bunnymatic:
problem
-------

tuning the styling around the main art piece and adding price and sold
tag if we have it.

solution
-------

* add `sold?` attribute to art piece
* if present show sold tag and strike-through on the price
* make art details slightly smaller font.
* add some spacing

screenshots
------------
image


kneyzberg:
very cool, love the sold tag, and the updated styling.

PR #138, opened 2023-12-12
Move ContentResource to ContentBlock

bunnymatic:
problem
-------

let's align on names.  `Resource` came from when we started with
DataMapper and everyone was calling it a resource.  but we don't care
anymore cuz ActiveRecord so...

solution
--------

rename ContentResource to ContentBlock which means table rename and a
few other clean ups.

NOTES
-----

This will motivate a `db:migrate` for both regular env and `RACK_ENV=test`


kneyzberg:
Understood, migrated on my end 👍

PR #141, opened 2023-12-17
Add common admin form styling

bunnymatic:
problem
--------

our forms are a mix of pico and some custom styling.  and are not
consistent.

solution
--------

add to `form_elements.scss` with some common form classes that we can
use to make sure they all style basically the same.

changes
------

* add form styling
* update classes to all forms so they pick up the styling
* move the last of the admin.js code that toggles the form visibility
  for exclusions page into a js class and add to globals
* update headers on admin pages with add new buttons to be more
      consistent so styling can be more compartmentalized.

screenshots
----------

**On this first one you can also see the header and "add new" button alignment which is consistent now on all index pages with "add new" plus sign icon**
Screen Shot 2023-12-17 at 5 35 12 PM
Screen Shot 2023-12-17 at 5 35 21 PM
Screen Shot 2023-12-17 at 5 35 48 PM
Screen Shot 2023-12-17 at 5 35 54 PM


frontend/stylesheets/admin/form_elements.scss
```diff
@@ -1,15 +1,41 @@
...
   }
 
   > a {
     display: block;
   }
 }
+
+form {
+  .form-input-field--label {
+    color: colors.$gray77;
```
bunnymatic, line 22: I tried to use the default `picocss` form styling by moving our elements to align with their DOM recommendation but it still didn't quite space out the way I wanted.  Partially because we have the `help` elements in there.  so...

added all this stuff.

frontend/stylesheets/admin/index.scss
```diff
@@ -1,11 +1,13 @@
...
-@import "./admin_main";
-@import './form_elements';
-@import './content_blocks';
-@import './image_resources';
-@import './palette';
-@import './env';
-@import './exclusions';
-@import './dashboard';
-@import './hero_images';
+@use './layout';
```
bunnymatic, line 2: I think we should leverage @use when we can.  this doesn't seem to break anything so... :D

views/admin/content_blocks/edit.slim
```diff
@@ -7,33 +7,39 @@
...
-        .lbl
-          label for=("content_block_#{ky}") = ky.capitalize
-        - if k == :body
-          textarea class="content_block" id="content_block_#{ky}" name="content_block[#{ky}]" rows=20 cols=75
-            = @content_block.send(ky) || ''
-        - else
-          input required=(k == :page) type='text' class="content_block" id="content_block_#{ky}" name="content_block[#{ky}]" value=(@content_block.send(ky) || '' )
-        span.help == help_msg[k]
+
+    .form-input-container
```
bunnymatic, line 15: just unrolled this form.   we don't need all the meta crap.  let's just be more explicit and easier to read/edit.

PR #142, opened 2023-12-17
Added React Bio component

kneyzberg:
### Problem 
Artist on Mau will submit a bio that is way too long and it will blow out the page. We need to create a react component that will only render a piece of the bio when it is too long.

### Solution 
- created a Bio_Box component to render the biotext.
- added conditions to render more/less functionality to collapse or expand bio.
- imported and used useBreakpoint to check for window size for conditional rendering.
- imported and used DOMPurify to sanitize HTML (just in case).
- added styling to react component.
- rendered component on artists/show page.
- updated zindexes to make sure component doesn't overlay nav and the menu bar.
- add bio box component to admin react test page.
- turn RenderHTML into its own general reusable component.
- add CharWord limit constant.

### To Dos (in following PRs)
- add in react testing library


frontend/javascript/react/components/bio_box/bio_box.tsx
```diff
@@ -0,0 +1,65 @@
...
+console.log("IN BIO BOX");
+interface BioBoxProps {
+  bioText: string,
+  numWords: number,
+};
+
+type BioBoxPlainProps = Pick<BioBoxProps, "bioText">
+type BioBoxExpandoProps = Pick<BioBoxProps, "bioText">
+
+const RenderHTML: FC<BioBoxPlainProps> = ({bioText}) => (<div dangerouslySetInnerHTML={{__html: DOMPurify.sanitize(bioText)}}/>)
```
bunnymatic, line 17: I like this.  Should it just be it's own component?  it feels like even though we might not want to use it that much, it could be helpful if we end up using react to show stuff from our CMS or the event data which will likely also be markdown that has to get marked up to HTML.

frontend/javascript/react/components/bio_box/bio_box.tsx
```diff
@@ -0,0 +1,65 @@
...
+      <button onClick={moreText} className="btn secondary bio-box--btn">{showMore ? "less" : "more"}</button>   
+    </div>
+  )
+}
+
+
+const BioBox: FC<BioBoxProps> = ({bioText, numWords})  => {
+  const { isMobile }= useBreakpoint()
+
+  const showWithExpando = !isMobile && numWords > 175
```
bunnymatic, line 53: Let's make this 175 into a named constant like `MAX_WORDS_BEFORE_EXPANDO` (terrible name) maybe `MAX_NUM_WORDS`... just so it's named and if we decide to tweak it we can see it easily.

frontend/javascript/react/hooks/use_breakpoint.tsx
```diff
@@ -0,0 +1,17 @@
+import vendorUseBreakpoint from 'use-breakpoint'
+
+
+const BREAKPOINTS = { mobile: 0, tablet: 768, desktop: 1280 }
```
bunnymatic, line 4: Can we change this to match the names and breakpoints from [pico](https://picocss.com/docs/containers.html)?
maybe something like
```
{ xs: 0, sm: 576, md: 768 <fill in the rest> }
```
then we can make sure if we use it elsewhere that breakpoints in our CSS that change stuff align with breakpoints here that also change stuff.

I think with this, you'll likely need to change the default to `xs` and then `isMobile = breakpoint === 'xs'` - or maybe we'll find that `sm` is the right breakpoint for this option.  I forget when we go down to 1 column.

package.json
```diff
@@ -17,5 +18,10 @@
   },
   "scripts": {
     "css-lint": "stylelint './frontend/stylesheets/**/*.{scss,css}'"
+  },
+  "dependencies": {
+    "dompurify": "^3.0.6",
```
bunnymatic, line 23: I think we don't need pico here - In fact I think this `pico` is not the same as `@picocss/pico` which we already have included above in the `devDependencies` list.  Or is it providing something useful?

Can we re-install these with the dev flag so they end up in dev dependencies?

I think it'd be something like
```
yarn add dompurify use-breakpoint --dev
```
you might have to `yarn remove` them first.

views/artists/show.slim
```diff
@@ -32,7 +32,8 @@
       - if @artist.studio_number.present?
         h3 = 'Studio ' + @artist.studio_number
       .bio--content.markdown
-        == markdown_content(@artist.bio)
+        .react-component
+          = react_component(id: 'bio_box', component:'BioBox', props: { bio_text: markdown_content(@artist.bio) , num_words: @artist.bio.split.count} )
```
bunnymatic, line 36: we could deal with this later but I wonder if we'd be better off using a markdown library in javascript instead.  if we did then we could maybe get around the `dangerously_set_html` thing.  Though it doesn't bother me and with the `dompurifier` it feels fine.  but... something we could think about later.  If we did, then maybe we could have - instead of your `RenderHTML` we have `RenderMarkdown` component which expects a markdown string and does all the conversion in there.  Not for this PR... just food for thought.

bunnymatic (commented):
We may need to update the z-index of this component or the header.  it looks like the bio is appearing on top of the header

image

Please add to the `zindex.css` file so we can keep track of all the indexes we're setting then you can `@use "zindex"` in your css and pull out the one you added

Also - if you want to leave one of these components on the admin test page, that is not the worst idea.  We don't have story book but if we just add any big components that we can to that page, then we also have sort of a little "index" of react components that we can use.

if you do that, make up some kind of division so it becomes like an index page. like

```
article.components-index
  h1 KittyCarousel
  == react_component...
article.components-index
  h1 BioBox
  == react_component ...
  
/ then we can fill in more samples below as we build them
```

bunnymatic:
Based on my first review, I think the things that should be addressed here are
* zindex issue
* breakpoints should match pico breakpoints
* move new yarn packages into devDependencies list

the rest could be done in later pr's.  either fast follow or we write some teeny clean up tickets and just put them in the list.

frontend/javascript/react/components/bio_box/bio_box.tsx
kneyzberg: Yeah I like the idea of it being a general component that we use everywhere, could be super helpful, happy to separate that out.

frontend/javascript/react/components/bio_box/bio_box.tsx
kneyzberg: For sure, makes sense!

frontend/javascript/react/hooks/use_breakpoint.tsx
kneyzberg: Yeah that makes sense, I'll take a look and match it up

package.json
kneyzberg: Yeah absolutely, i was wondering why they wouldn't show up in dev, i'll fix and make sure they end up there.

views/artists/show.slim
kneyzberg: I think that could be useful/helpful, definitely something to discuss

frontend/stylesheets/zindex.scss
```diff
@@ -4,4 +4,6 @@
  *
  */
 $mobile-menu: 8;
```
bunnymatic, line 6: minor - let's put these in order - bottom to top maybe?  so we can easily see who's on top of whom.

bunnymatic (approved):
Looks good.  I just put that prettier/eslint branch in which is leading to the conflicts.  hopefully not too painful to unpack.

frontend/stylesheets/zindex.scss
kneyzberg: for sure, on it!

PR #144, opened 2023-12-18
Add prettier and eslint

bunnymatic:
problem
--------

javascript is sloppy

solution
-------

add prettier and eslint so at least we could run

```
yarn eslint --fix
yarn prettier --write
```

and clean things up.

we might later put this in a CI task and enforce it.  but for now, this
is just to get the tooling in there.


bunnymatic:
Note: you can also `yarn tsc` now to get a typescript check. It doesn't pass cuz we've been sloppy but at some point we might start to enforce some of that too.

PR #149, opened 2023-12-23
Add new eventing and show it on the main page

bunnymatic:
problem
---------

we want to support scheduled events that can easily and quickly control
the content that lives on the front page.  this includes the banner
image and the content.

solution
--------

build it!

The new model StudioEvent has 2 ImageResource references, banner and
display, a headline, description and active date range.  With this
change we can throw out the ImageResource admin views because it should
become more of a utility not a place to just add images.

changes
-------

* Add StudioEvent model
* Add connected controllers and repos and views
* Update front page to pull active studio event data if it exists then
  falls back to CMS then to default.
* remove hero images as we don't need them anymore
* remove admin view on ImageResource since we don't need admin control
* update form elements styling to better handle an image file form
* added friendly id for slug management
* added rspec-collection_matchers
* updated admin `add` buttons to `Add...` in there label

screenshots
-----------

Screenshots in order:
* Default front page
* Go to studios index page
* Edit the event to be active (set activate_date to now)
* go back to the home page and see the new content

Screen Shot 2023-12-23 at 9 20 36 AM
Screen Shot 2023-12-23 at 9 21 08 AM
Screen Shot 2023-12-23 at 9 21 29 AM
Screen Shot 2023-12-23 at 9 21 46 AM

upcoming work 
---------------

* add a way to preview the home page if you're authorized for an event that is not yet active
* add events list and show pages for the public to see all our events


controllers/admin/image_resources_controller.rb
```diff
@@ -29,12 +29,5 @@ class App < Sinatra::Base
       Repos::ImageResourceRepo.save(filename, tempfile)
       redirect '/admin/image_resources'
     end
-
-    ### delete
-    get '/admin/image_resources/:slug/delete' do
```
bunnymatic, line 34: we may soon just get rid of all the image resources controller/views - didn't quite want to dump them yet but we removed `slug` and this method depended on slug so it was just easier to dump it.

db/schema.rb
```diff
@@ -27,22 +27,27 @@
...
     t.integer "width", null: false
     t.integer "height", null: false
     t.bigint "file_size", null: false
     t.binary "image", null: false
-    t.index ["slug"], name: "index_image_resources_on_slug", unique: true
+  end
+
+  create_table "studio_events", force: :cascade do |t|
+    t.string "slug", null: false
+    t.bigint "banner_image_id"
```
bunnymatic, line 42: Currently the frontend won't let this be optional but it feels like maybe we'd want an event that doesn't have a banner so leaving this db attr nil-able and if we want to we can loosen the frontend.

repos/image_resource_repo.rb
```diff
@@ -9,8 +9,8 @@ def self.all
       ImageResource.all
     end
 
-    def self.destroy(slug)
-      ImageResource.find_by(slug:).destroy
+    def self.destroy(resource)
```
bunnymatic, line 12: maybe something to discuss as a concept.  this repo returns AR records.  Should it take AR records as inputs?  or should it be more agnostic and take only ids/slugs?


This change was because we dumped slug.  I could have sent ID in here but I didn't for some reason...

spec/features/admin/admin_routes_spec.rb
```diff
@@ -8,7 +8,19 @@
   end
 
   context 'Admin pages:' do
-    %i[image_resources/new image_resources content_block content_blocks env exclusions palette].each do |endpoint|
+    context 'Protected endpoints:' do
```
bunnymatic, line 11: moved this from `app_spec` cuz it's more about what's unauthorized and related to `admin_routes`

PR #185, opened 2023-12-29
Implemented ArtModal on Test Admin page

kneyzberg:
### Problem 
As part of the user experience on our website, when a user will click on a photo on and Artist's page, it should open up a popup that will allow you to scroll through that Artist's Art Pieces. 

### Solution 
Implement a ArtCarousel using react-modal that allows you to click through artist work using buttons. 

- created ArtModal, ArtCarousel, ArtWindow, ArtPieceInfo, and Button components for the carousel.
- pulled in useCarouselState and useModalState from Mau in order to manage state properly for modal.
- added styling to react components to style ArtWindow.
- added reader 'model' for model data on ArtPiece model.
- updated filenames for react hook files that are just regular typescript.
- broke up components into various files for clarity and separation of concerns.
- added testing for all art_modal components.


frontend/javascript/react/components/art_carousel/art_modal.tsx
```diff
@@ -0,0 +1,44 @@
...
+type ArtPiece = any;
+
+interface ArtModalProps {
+  artPieces: ArtPiece[];
+}
+
+const ArtModal: FC<ArtModalProps> = ({ artPieces }) => {
+  const { isOpen, open, close } = useModalState();
+
+  const customStyles = {
```
bunnymatic, line 18: does this have to live outside of CSS?  I only ask to reduce the number of places we need to think about where we find styling.  If we can put this in a CSS, file like our other components, let's do it.  If it's not convenient because of the way the modal thing works, then let's just make a note.  

Kind of related - this zIndex hard coded here defeats or is disconnected from our `zindex.scss` list which is going to make it a little dangerous.  If we just decided to multiple all zindex's by 10 to give us more space, this is not visible and would likely lead to some issues.  Food for thought.

frontend/javascript/react/components/art_carousel/art_window.tsx
```diff
@@ -0,0 +1,78 @@
...
+        <div className="art-window-image" style={imageStyle}></div>
+      </div>
+      <div className="art-window-info--container">
+        <div className="art-window-info--left">
+          <ArtPieceInfo
+            label="Title: "
+            value={artPiece.title}
+            classes="art-window-info--title"
+          />
+          <ArtPieceInfo
```
bunnymatic, line 50: super nit --- common indentation here and through this file.  prettier might just fix all that with `yarn prettier --write`

you have
```
<ArtPieceInfo
  label
/>

// and

<ArtPieceInfo
label
/>
```

frontend/javascript/react/components/art_carousel/art_window_buttons.tsx
```diff
@@ -0,0 +1,43 @@
+import React, { FC } from "react";
+import "./art_modal.scss";
+
+
+interface PrevButtonProps {
+  previous(): number,
```
bunnymatic, line 6: `previous` is a function that returns a number?  this type definition looks weird to me,.
if you `yarn tsc` do you get a lot of whining?

i think you might mean
```
previous: () => void;
```

frontend/javascript/react/components/art_carousel/art_window_buttons.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+  return (
+    <a href="#" className="secondary art-window--prev" onClick={(e) => {e.preventDefault(); previous()}}>
+      <i className="icon-circle-left art-window--buttons-bg"></i>
+    </a>
+  )
+}
+
+interface NextButtonProps {
+  next(): number,
```
bunnymatic, line 19: same as above - i think this is an invalid interface definition.  Clearly our typescript guards/tooling are pretty light :D

frontend/javascript/react/components/art_carousel/art_window_buttons.tsx
```diff
@@ -0,0 +1,43 @@
...
+}
+
+const NextButton: FC<NextButtonProps> = ({next}) => {
+
+  return (
+    <a href="#" className="secondary art-window--next" onClick={(e) => {e.preventDefault(); next()}}>
```
bunnymatic, line 25: I like these teeny components. And it may seem a little extra but could we pull out this prevent default?  it's hard to see/read when it's all jammed in the onClick.

maybe like
```
const NextButton: FC<NextButtonProps> = ({next}) => {
  const handleClick = (e) => {
    e.prevent_default();
    next();
  }

  return <a onClick={handleClick} ...
```

Kind of makes me wonder if we need a super base component which is something like LinkThatActsLikeButton that would do the prevent default for us?  or a `LinkButton` which is actually a button under the hood and we just strip off all the styles so it looks like text.   Not for this PR... just thinking

bunnymatic, line 25: lets' try to be accessible - add sensible `aria-label` and `title` attributes to all these links.

frontend/javascript/react/components/art_carousel/art_window_buttons.tsx
```diff
@@ -0,0 +1,43 @@
...
+}
+
+interface CloseButtonProps {
+  close(): any;
+}
+
+const CloseButton: FC<CloseButtonProps> = ({close}) => {
+  return (
+    <a href="#" className="secondary art-window--close" onClick={(e) => {e.preventDefault(); close()}}>
+      <i className="icon-cross"></i>
```
bunnymatic, line 38: Could we leverage my `Icon` component :D

```
<Icon name='cross'/>
```

frontend/javascript/react/hooks/use_carousel_state.test.ts
```diff
@@ -0,0 +1,175 @@
+import { describe, expect, it } from "@jest/globals";
+import { act, renderHook } from "@testing-library/react-hooks";
```
bunnymatic, line 2: Did this work?  I think @testing-library/react might have renderHook so we might not need to add this library.  not sure.

frontend/stylesheets/colors.scss
```diff
@@ -3,6 +3,7 @@
 $white: #fff;
 $black: #1b1612;
 $dark-gray: #2b1612;
+$dark-slate-gray: #3D475C;
```
bunnymatic, line 6: we should align on our naming strategies.  without a designer it's tougher but... 
these'll totally fly for now.

frontend/stylesheets/mixins.scss
```diff
@@ -1,5 +1,44 @@
...
+
 @mixin ellipsis {
   overflow: hidden;
   text-overflow: ellipsis;
   white-space: nowrap;
 }
+
+
+@mixin light-border {
+  border: 1px solid lighten(#5a636a, 47%);
```
bunnymatic, line 11: can we put this color in our palette or pull something from our palette that is close?

frontend/stylesheets/mixins.scss
```diff
@@ -1,5 +1,44 @@
...
+
+@mixin flex-center-center {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+}
+
+@mixin floating-window-border {
+  @include light-border;
+  box-shadow: 0px 0px 8px rgba(200, 200, 200, 0.8);
```
bunnymatic, line 22: can we use a color from our palette for this gray?

views/admin/test/index.slim
```diff
@@ -5,9 +5,11 @@ ruby:
...
 
-  numWords = bioText.split.count
+  num_words = bio_text.split.count
+
+  art_pieces = Repos::ArtistRepo.all.first.art_pieces.map(&:model)
```
bunnymatic, line 12: I wonder if - for the test page, we could have less dependency on actual data being present?  we could manually just construct some "art pieces" because the thing the react component gets is just a hash, right?

what about something like
```
  art_pieces = images.map.with_index do |image, idx|
    {
      title: "Untitled #{idx}",
      year: rand(2000..2020),
      sold: !!(idx % 2),
      price: "$#{rand(1..200)*100}",
      medium: ["stuff", "paint", "paper"].sample,
      image_urls: {
        medium: image,
        large: image,
        original: image,
      }
    }
  end
  ```
  instead of calling the artist repo?

File: frontend/javascript/react/components/art_carousel/art_modal.tsx [hunk 12]
```diff
@@ -0,0 +1,44 @@
+import React, { FC } from "react";
+
+import { useModalState } from "../../hooks/use_modal_state";
+import Modal from "react-modal";
+import { ArtCarousel } from "./art_carousel";
+import "./art_modal.scss";
+
+
+type ArtPiece = any;
```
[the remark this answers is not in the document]
kneyzberg: Copy that, I can start that directory.

frontend/javascript/react/components/art_carousel/art_modal.tsx
kneyzberg: I don't think it has to, I think that's just how react-modal sets it up, but I'm sure we can move into a css class.

frontend/javascript/react/components/art_carousel/art_window.tsx
kneyzberg: whoops, will fix!

frontend/javascript/react/components/art_carousel/art_window_buttons.tsx
kneyzberg: Sure, I wasn't sure exactly how to write these, did some googling and maybe did not come back with the right thing, will fix.

frontend/javascript/react/components/art_carousel/art_window_buttons.tsx
kneyzberg: Sure thing, I think that totally makes sense, and we can talk about that in a follow up or something.

frontend/javascript/react/components/art_carousel/art_window_buttons.tsx
kneyzberg: Absolutely!

frontend/javascript/react/hooks/use_carousel_state.test.ts
kneyzberg: It's still not working, but it's more from the fact that `let result; ` has an implicit type of any that is making it upset.

frontend/stylesheets/mixins.scss
kneyzberg: sure thing!

frontend/stylesheets/mixins.scss
kneyzberg: I'm sure we can, i was just pattern matching to MAU

views/admin/test/index.slim
kneyzberg: Yeah i'm sure we can do that, to avoid having to call the Repo.

frontend/javascript/react/components/art_carousel/art_modal.tsx
kneyzberg: Jk, we do need this, this is for the overlay of how React Modal works, but i can try and look into how to finagle it into a css class instead.

kneyzberg: I think it's possible but it's not playing nicely when I try to do it, i'm going to table for now, but maybe we can pair on it to get it working the way we need.

frontend/javascript/react/hooks/use_carousel_state.test.ts
bunnymatic: with some recent mods I think the `import { describe, expect }` are going to cause issues too so should probably dump that line.  

And with #187 builds are going to break if the js tests break. Feel free to just rename the file to ._testjsx or something that doesn't get picked up by jest or comment things out or whatever.  we can dig on how get hooks tested in a separate pr.

views/admin/test/index.slim
bunnymatic: I think the bigger win is that if for some reason (when I'm testing this on a machine that doesn't have MAU handy), I can still see this page and the example.  

Also it'll give is way more control about properties so we can update it to verify that price doesn't show if there's no price or whatever.

frontend/javascript/react/hooks/use_carousel_state.test.ts
kneyzberg: Sounds good 👍

views/admin/test/index.slim
kneyzberg: Yeah this makes complete sense,  working on implementing this now!

frontend/javascript/react/components/art_carousel/art_window.tsx
```diff
@@ -0,0 +1,75 @@
...
+            value={artPiece.year}
+            classes="art-window-info--year"
+          />
+        </div>
+      </div>
+    </div>
+  );
+};
+
+export { ArtWindow, ArtPieceInfo, ArtPieceInfoProps, ArtWindowProps };
```
bunnymatic, line 75: we should only export what we need.  Since `ArtPieceInfo` is kind of a private class helper... we could likely not export that stuff.

I know that we are exporting types for tests (which I kind of don't like but...).

food for thought.

frontend/javascript/react/components/icon.tsx
```diff
@@ -34,11 +35,12 @@ type IconName =
 
 interface IconProps {
   name: IconName;
+  classes?: string;
 }
 
-const Icon: FC<IconProps> = ({ name }) => {
+const Icon: FC<IconProps> = ({ name, classes }) => {
```
bunnymatic, line 41: how are we using classes here?  do we need to decorate or icons more than this component supports?

frontend/javascript/react/components/types/interfaces.ts
```diff
@@ -0,0 +1,17 @@
+interface ArtPiece {
+  artistId?: number;
+  artistName: string;
+  dimensions?: string;
+  displayPrice?: string;
+  id: string;
+  imageUrls: Record<string, String>;
```
bunnymatic, line 7: we could even type this thing.  but this is a great start.

I wonder if this should be
`types/art_piece`

and then we have
`types/index.ts` that just imports all our `types` files...

PR #186, opened 2023-12-30
Add selenium so we can do full JS browser testing

bunnymatic:
problem
--------

Running full e2e tests now that we're doing more react is important.  We
need our browser to have JS support.  But the current Capybara setup
just uses the `rack_test` driver which does not have JS support.

solution
--------

This PR introduces selenium and webdrivers gems and sets up capybara so
that we can tag certain tests with `type: :e2e` and they will run
selenium and a browser.  It also includes support for headed or headless
chrome for testing.  By default it's headed for development so we'll see
the tests run.  If this turns out to be too annoying we can switch it
later.

changes
------

* add e2e spec around editable content for events/cms entries
* split css and js entries
* move from `thin` to `puma`
* add tests controller and sample data create/delete endpoints
* revamp admin js and application js so that they were correct for our
  test runner
* reinstate content block edits and fix that so it works.

notes
------

I spent a lot of time trying to get the database transaction issue
figured out with no luck.  The basic issue is that in the rspec example
or it block, if you use factory bot to create database entries, that
data is in a transaction which is separate from the db connection that
the app is using which means the application does *not* see that
creation.  Rspec-Rails and some other rails stuff has tooling to get
around this but I was unable to figure out how to leverage that here or
mimic it.

Because of this, I added a tests_controllers.rb file and a
`tests/sample_data` endpoint that *only* works if you're in test mode
(for safety).

When we write e2e tests we should populate this sample data such that it
is a valid start up scenario for *all* e2e tests.  We can think of it as
a set of seed data for a test instance of the application.  Currently,
we're going to start with 1 active event and 2 cms entries for the two
pages that show custom cms content.  I could imagine a default
"exclusion" item... in the future.

It's not great but it's better than not having full feature e2e tests.

Exercise for the author and reader
-------------------

I don't understand `@use` vs `@import` in SCSS and I feel like I should.


controllers/tests_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+    # I'd rather have tests, so... this endpoint can be called from
+    # a test suite to do some simple FactoryBot creation of things.
+    #
+    # Right now it creates a studio event and cms content for
+    # event main and home main.  Any more control of this needs to be
+    # managed via navigating the application.
+    #
+    # If we find it useful, we might add some params that could help
+    # have more control over the objects.
+    get '/tests/sample_data' do
```
bunnymatic, line 21: Though I don't totally like this, it's interesting to think about because it *would* likely be the kind of solution we'd need if we used `cypress` or `playwright` which, as e2e solutions, are way outside of the `application` and would not have any way to control factory bot or other data creation for tests.  So maybe it's not as bad as I initially thought.

frontend/entrypoints/admin.js
```diff
@@ -5,6 +5,3 @@
 console.log("Vite ⚡️ Ruby");
 
 import "../javascript/admin.js";
-
```
bunnymatic, line 8: Split out CSS because when we were building in test mode, the browser was not getting both js and css so... even though vite suggests it's supposed to generate both tags even if they're in a single tag... let's just keep it simple and keep the css with a `vite_stylesheet_tag` and js `vite_javascript_tag`.

This also motivated the move from `@use` to `@import` to make things work - I still don't understand the difference.  And need to learn.

README.md
```diff
@@ -146,6 +146,24 @@ To run the spec tests,
...
+brew install chromedriver
+```
+
+This tends to fall out of date so if you run into issues, you may need to
+
+```
+brew upgrade chromedriver
+```
+
+Also because of Mac software protections, after an install or upgrade you need to (one time) open the app manually to get Apple to think the software is not malicious.  To do this open the folder it's in with Finder (use `which chromedriver` if you don't know - might be `/usr/local/bin`).  Then hold down option and double click the app.  This will prompt you to "open" and accept the developer details.  After that you should be good until the next upgrade.
```
bunnymatic, line 165: Let's add whatever env settings we need to make it run headless - maybe
```
CI=true bin/rspec
```

kneyzberg: copy that!

kneyzberg (approved):
Confirmed that checking out the branch, installing, and running the tests works as expected! Everything passes 👍

PR #209, opened 2024-01-04
Created ArtCard Component and Implemented Component on Artist Page

kneyzberg:
### Problem
In order to implement our ArtModal on the Artist Page, our trigger for opening the modal needs to be a react component. 

### Solution
Replace the slim partial `_artpiece_card.slim` with a react component called ArtCard. 

- create ArtCard Component to match slim partial of `_artpiece_card.slim`
- ported over card styling
- added ArtCard to test admin page and replace slim partial with react component
- added testing


frontend/javascript/admin/section_toggler.ts
```diff
@@ -6,13 +6,15 @@ class SectionToggler {
...
       const sections = document.querySelectorAll(sectionToToggle);
       for (const section of sections) {
-        const hidden = section.attributes.getNamedItem('aria-hidden') || section.attributes.getNamedItem('hidden');
+        const hidden =
+          section.attributes.getNamedItem("aria-hidden") ||
+          section.attributes.getNamedItem("hidden");
         if (hidden) {
-          section.removeAttribute('hidden')
-          section.removeAttribute('aria-hidden')
+          section.removeAttribute("hidden");
```
bunnymatic, line 13: i wonder if i missed this or if our prettier/eslint configs are not aligned.  not a big deal here just curious.

frontend/javascript/react/components/art_card/art_card.tsx
```diff
@@ -0,0 +1,25 @@
...
+import * as types from "../types";
+
+interface ArtCardProps {
+  artPiece: types.ArtPiece;
+}
+
+const ArtCard: FC<ArtCardProps> = ({ artPiece }) => {
+  return (
+    <div className="artpiece-card">
+      <a href="#">
```
bunnymatic, line 14: should we dump this link?  if we want it clickable we can add it back in whenever we make it clickable.

spec/features/artists_pages_spec.rb
```diff
@@ -51,23 +51,7 @@
       artwork_heading = find('.artist-page--artwork--title h2')
       expect(artwork_heading.text).to match 'Artwork'
       expect(page).to have_css('.artpieces-grid')
-      expect(page).to have_css('.artpiece-grid--item .artpiece_card--image')
-    end
-
-    it 'renders thumb artist image as main image on artist page visit with query params' do
```
bunnymatic, line 57: not totally sure but you could try `type: :e2e` with this test and see if it still works.  
```
it 'renders ...', type: :e2e  do
```

might be nice to have some kind of test that still proves out "we see the artwork"

bunnymatic (approved):

frontend/javascript/admin/section_toggler.ts
kneyzberg: That is great question, maybe we can sanity check this weekend, I just ran yarn-lint-fix

frontend/javascript/react/components/art_card/art_card.tsx
kneyzberg: Isn't the trigger going to be clicking this art-card/any part of it? Is there are a better way to instruct to the user that it's clickable than a link?

spec/features/artists_pages_spec.rb
kneyzberg: Copy that, let me see if i can bring it back in

frontend/javascript/react/components/art_card/art_card.tsx
bunnymatic: it is "going to be" but it's not now.
the idea of putting functionality in before it's functional is what I don't like.

I know it's going to be a super fast follow... but :)
if you quit right now, then this `a` could be misleading.   but it would be perfectly functional without it.

it's a super process-y nit.  we can discuss.  totally fine with the merge.

spec/features/artists_pages_spec.rb
```diff
@@ -53,6 +53,23 @@
       expect(page).to have_css('.artpieces-grid')
       expect(page).to have_css('.artpiece-grid--item')
     end
+
+    it 'renders thumb artist image as main image on artist page visit with query params',
```
bunnymatic, line 57: 👍

frontend/javascript/react/components/art_card/art_card.tsx
kneyzberg: I see, no that totally makes!

PR #210, opened 2024-01-05
Implemented ArtModal on Artist Page

kneyzberg:
### Problem 
Now that we have ArtCard created as a React component, we need to implement the Art Modal and set the ArtCard as the trigger for opening the Modal on the Artist `show` page. 

### Solution
- separated out ArtModal into ArtModalModal and ArtModalTrigger
- created ArtBrowser wrapper component for artpieces on Artist Page
- ported over styling to react component
- created ArtPieceContext to keep track of which artPiece is clicked

### To Dos
- add in 'medium' attribute into the data the react component receives


frontend/javascript/react/components/artwork_browser/artwork_browser.tsx
```diff
@@ -0,0 +1,39 @@
...
+  const [currentIdx, setCurrentIdx] = useState(0);
+
+  return (
+    <ArtPieceContext.Provider value={currentIdx}>
+      <div className="artwork-grid">
+        {artPieces.map((artPiece, idx) => (
+          <div key={idx} className="artwork-grid--item">
+            <ArtCard
+              open={open}
+              setCurrentIdx={() => setCurrentIdx(idx)}
```
bunnymatic, line 28: this use of `setCurrentIndex` feels weird to me sort of.

I think it's because the `ArtCard` shouldn't really know about what "current" means.  We could put that art card anywhere on the site and "currentIndex" only makes sense if we are inside this ArtworkBrowser.  So... then I wonder - why do we need current index here?  if we need it, could we rename this prop so that it indicates what it's for and leaves ArtCard less smart about where it might live.  

similarly, I wonder about `open`?  again, ArtCard doesn't maybe know it's in a thing that can open a modal.  Could it instead be `onClick` so then we could make it do anything.  And in this one case we decide it's `open the modal`.

like
```
<ArtCard onClick={open} ...
```

frontend/javascript/react/components/artwork_browser/artwork_browser.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+const ArtPieceContext = createContext(0);
+
+const ArtworkBrowser: FC<ArtworkBrowserProps> = ({ artPieces }) => {
+  const { open, close, isOpen } = useModalState();
+
+  const [currentIdx, setCurrentIdx] = useState(0);
+
+  return (
+    <ArtPieceContext.Provider value={currentIdx}>
```
bunnymatic, line 22: nice.  if we're going to put provider to manage current index, feels like it should also provide the setter.  maybe?

frontend/javascript/react/components/artwork_browser/artwork_browser.tsx
```diff
@@ -0,0 +1,39 @@
...
+      <div className="artwork-grid">
+        {artPieces.map((artPiece, idx) => (
+          <div key={idx} className="artwork-grid--item">
```
bunnymatic, line 25: some react linters will not let you get away with using `idx` here because someone could sort the array and then the key's don't *actually* align with the object and re-rendering *can* get confused.  Could we use `artPiece.id`?

frontend/javascript/react/components/artwork_browser/artwork_browser.tsx
bunnymatic, line 22: Looks like you haven't checked in the provider, but when you do, let's wrap it all up nicely so no-one outside needs to know that `.Provider` is a thing.  If you check out `art_pieces.context.tsx` in MAU, you'll see we both define/create the context and export it and make a provider and export it. and all of that lives in 1 file that is "the context".  

```
export const ArtPiecesContext = createContext<types.ArtPiecesContext>({});
export const ArtPiecesProvider: FC<ArtPiecesProviderProps> = ({
  children,
  dataForTheContext,
}) => {
  return (
    <ArtPiecesContext.Provider value={{ dataForTheContext }}>
      {children}
    </ArtPiecesContext.Provider>
  );
};
```

Its just a little compartmentalization.    In some other project, we further subclassed `createContext` to include some more error checking and stuff so if you follow this kind of code isolation, it's easier to apply those kinds of changes to the whole app as opposed to trying to find everywhere we did `.Provider` etc.

I noticed we did *not* do the following in MAU but now I might take it even one step further which would be to make a teeny little hook to wrap the useContext into something like `useArtPieceContext` .  I think it'd look something like

```
const useArtPieceContext = () => useContext<ArtPieceContext>(initialValue)
```
not sure if that's exactly right but...
not required but food for thought.

frontend/javascript/react/components/artwork_browser/artwork_browser.tsx
```diff
@@ -0,0 +1,39 @@
...
+interface ArtworkBrowserProps {
+  artPieces: types.ArtPiece[];
+}
+
+const ArtPieceContext = createContext(0);
+
+const ArtworkBrowser: FC<ArtworkBrowserProps> = ({ artPieces }) => {
+  const { open, close, isOpen } = useModalState();
+
+  const [currentIdx, setCurrentIdx] = useState(0);
```
bunnymatic, line 19: typing - `useState<number>(0)`

bunnymatic (commented):
I realize you're not done yet but... left thoughts anyway :)

Looking really good.

it immediately makes me want to rejigger the artist page design so that the images are smaller and we lean hard on this carousel - not for this PR - just thinking.

frontend/javascript/react/components/art_carousel/art_carousel.tsx
```diff
@@ -13,9 +14,11 @@ interface ArtCarouselProps {
 }
 
 const ArtCarousel: FC<ArtCarouselProps> = ({ artPieces, close }) => {
+  const currentPieceIdx = useContext(ArtPieceContext);
+
   const { current, next, previous } = useCarouselState<ArtPiece>(
     artPieces,
-    artPieces[0],
+    artPieces[currentPieceIdx],
```
bunnymatic, line 21: after looking at this - I wonder - what's the point of a context to store current when this useCarouselState is keeping track of current?  are these doing the same thing?   I mean I know they are not but...
Don't we only need to tell the carousel "start here"?  so I wonder if `ArtModalModal` should get `artPieces` and `initialIndex` or `initial`.  then let that pass that value down through the carousel (or use a context to init).  

We might think about a slight word change.  If the carousel uses `current` which is the current one you're looking at, maybe the other things that kick things off might use the word `initial` instead because as soon as the user clicks next, that 'currentPieceIdx' won't be current any more.  And then I wonder if that language might help us understand where we need to set "initial".  


I see why it is a little confusing.   Sometimes that confusion means a code smell.  If we wrote the code and we can't understand it then... what are we doing :D ?!?

app/controllers/artists_controller.rb
```diff
@@ -27,7 +27,7 @@ class App < Sinatra::Base
                           @artist.art_pieces.first
                         end
 
-      @secondary_art_pieces = @artist.art_pieces.reject { |ap| ap.id == @main_art_piece.id }
+      @secondary_art_pieces = @artist.art_pieces.reject { |ap| ap.id == @main_art_piece.id }.map(&:model)
```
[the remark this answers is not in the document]
kneyzberg: Yes I was thinking the exact same thing!

frontend/javascript/react/components/art_carousel/art_modal_modal.tsx
```diff
@@ -0,0 +1,55 @@
...
+import * as types from "../types";
+import { ArtCarousel } from "./art_carousel";
+
+interface ArtModalModalProps {
+  artPieces: types.ArtPiece[];
+  isOpen: boolean;
+  close: () => void;
+}
+
+interface ArtTiggerProps {
```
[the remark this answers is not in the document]
kneyzberg: Whoops, will fix!

frontend/javascript/react/components/artwork_browser/artwork_browser.tsx
kneyzberg: Yeah I was thinking about this, I don't love it and don't love having to pass down a SetState function to what should be sort of a dummy component? But I understand what you're saying and could modify to be onClick

frontend/javascript/react/components/artwork_browser/artwork_browser.tsx
kneyzberg: Sure, It's been a while since I've worked with useContext, so maybe I might need some walkthrough on how to do this right.

frontend/javascript/react/components/artwork_browser/artwork_browser.tsx
kneyzberg: Yes, absolutely, I forgot about this

frontend/javascript/react/components/artwork_browser/artwork_browser.tsx
kneyzberg: Sure, I was looking at this, and can definitely do that!

frontend/javascript/react/components/art_carousel/art_carousel.tsx
kneyzberg: Yeah you're correct in that we only need to tell the carousel start here. My question would be how does ArtModalModal know what index to pass through without the click of the ArtCard, is there a better to notify that initial index?

Ahh i see, yeah changing the words up would be helpful

frontend/javascript/react/components/art_card/art_card.tsx
```diff
@@ -6,12 +6,20 @@ import * as types from "../types";
 
 interface ArtCardProps {
   artPiece: types.ArtPiece;
+  open: () => void;
+  setCurrentIdx: () => void;
 }
 
-const ArtCard: FC<ArtCardProps> = ({ artPiece }) => {
+const ArtCard: FC<ArtCardProps> = ({ artPiece, open, setCurrentIdx }) => {
```
bunnymatic, line 13: I think this refactor actually exposes how we can make this component dumber and push the control of the "current" out of here.

if the props here were

```
interface ArtCardProps = { artPIece: ArtPiece, onClick: () => void }
```
then this component just is
```
const handleClick: MouseEventHandler = (e) => {
   e.preventDefault();
   onClick()
};

<div>
   <a href="#" onClick={handleClick}> 
     ...
```

then in the parent we do something like

```
   const handleCardClick: (index: number) => () => void = ({index}) => {
       return () => {
            setInitialIndex(index)
            open()
       }
   }
  
   ... in the render 
 
   { artPieces.map( ({artPiece, idx}) => (<ArtCard artPiece={artPiece} onClick={handleCardClick(index)} />) }
```

this way we're closing over the "index" by building custom handler functions with the index built into them which we can hand down and now the Card doesn't know anything about `initial` or modal `open`.  it just knows, if I click on the card, I run the function you gave me.

And I bet with this change, the need for the context might just go away because it's all managed in one place.  not totally sure but maybe.

bunnymatic:
Totally good news - it looks good on mobile

image

bunnymatic:
for a follow up pr, let's make the modal window white and the overlay darker like `rgba(50,00,60,0.4)` or something - dark purple transparent... :D

frontend/javascript/react/components/art_card/art_card.tsx
kneyzberg: This makes sense, I'm implementing this and it totally works. I just pass the value down through ArtModalModal, to Carousel and it works fine!

PR #211, opened 2024-01-07
Implemented webhook to account for artist updates on MAU

kneyzberg:
### Problem
We want our site to be notified whenever an Artist makes and update to their information so that we can update the cache and refresh the Artist information right after the update happens. 

### Solution
Implement a webhook that listens for post requests from Mau and refreshes the cache based on the updated artist's id. 

- created webhook helper that checks if request being made is coming from mau w/ correct auth headers.
- Added webhook controller for webhook endpoint. 
- Added method on Artist Repo to refresh/artist if endpoint is hit.
- updated safe_cache to have a delete method by key.
- added specs for webhook controller and artist repo.


bunnymatic (commented):
looks good.  we need to coordinate with a little bit with https://github.com/bunnymatic/mau/pull/393

bunnymatic (approved):

PR #212, opened 2024-01-08
Update our jest test `render` helper to use default props

bunnymatic:
problem
-------

we often run jest tests with the same basic props.  but our current
pattern doesn't easily support that.

solution
--------

update tests helper `r` methods to take optional `propsOverrides` which
is only the props we need or may override in tests, and then let the
normal `r()` behavior just render a component with the default props.


frontend/javascript/react/components/art_carousel/art_modal_modal.test.tsx
```diff
@@ -32,38 +29,29 @@ describe("ArtModalModal", () => {
...
-      initialIdx: 0,
-      close: closeMock,
     });
 
     expect(queryByText("Test Art Piece 1")).not.toBeInTheDocument();
   });
 });
-
-describe("ArtModalTrigger", () => {
-  const r = (props: ArtModalTriggerProps) =>
```
bunnymatic, line 59: while I was here - I dumped this `ArtModalTrigger` because it's only used in tests.  we can now put anything in as a trigger so... let's let it be open ended.

kneyzberg: Sounds good, makes sense 👍

kneyzberg:
Awesome, this cleans things up nicely!

PR #231, opened 2024-01-13
Don't include css with dev server

bunnymatic:
problem
-----

there are styling include issues.

this doesn't solve them but it's closer.

solution
-------

make sure with dev server running we *don't* include the css cuz vite is
supposed to take care of that.

notes
----

we're still not seeing the font so `pico` is winning.
also it feels like there are some other overrides that shouldn't be
happening.  i think this is better than we were but still more to do.


frontend/entrypoints/admin.js
```diff
@@ -4,4 +4,5 @@
 // and link to the scripts and stylesheets, and let Vite transform it.
 console.log("Vite ⚡️ Ruby");
 
+import "../stylesheets/admin/index.scss";
```
bunnymatic, line 7: maybe these are in the wrong order?!?

PR #232, opened 2024-01-13
Added Keyboard Handlers for Arrow Keys in Art Carousel

kneyzberg:
### Problem 
When the art modal is opened, we can navigate by clicking, but it would beneficial to also allow navigation through the carousel by pressing the left and right arrow keys. 

### Solution 
- Add Keyboard Event Listener and Handler to listen for key presses on ArrowLeft and ArrowRight to trigger navigation through the carousel.


frontend/javascript/react/components/art_carousel/art_carousel.tsx
```diff
@@ -23,6 +25,18 @@ const ArtCarousel: FC<ArtCarouselProps> = ({
     artPieces[initialIdx],
   );
 
+  const keyDownHandler = useCallback((e: KeyboardEvent) => {
```
bunnymatic, line 28: useCallback should have a dependency array
```
const handler = useCallback((e: KeyboardEvent) => {
}, [ previous, next ]);
```

bunnymatic: Is KeyboardEvent an enum?  I wonder if we could do
```
if e === KeyboardEvent.ArrowLeft
```
so we wouldn't need these loose hard coded strings here?
Do you know where `KeyboardEvent` is coming from?  I don't see an import here for that type. Maybe it's just a kind of native.

If that didn't work, you could maybe export constants from `useEventListener` so that typescript will help us not make mistakes.

```
// in use_event_listener

export const EVENT_KEYS = {
  ArrowLeft: "ArrowLeft",
  ArrowRight: "ArrowRight",
};


// here
   if (e.key == EVENT_KEYS.ArrowLeft) {
   
```
It might seem a little silly but then you can't do `EVENT_KEYS.ArowLeft` without someone complaining about the typo.

frontend/javascript/react/components/art_carousel/art_window_buttons.tsx
```diff
@@ -1,7 +1,6 @@
 import "./art_carousel.scss";
 
-import React, { FC } from "react";
-
+import React, { FC , useCallback} from "react";
 import { Icon } from "../icon";
```
bunnymatic, line 4: don't need `useCallback` import here.

frontend/javascript/react/hooks/use_event_listener.js
```diff
@@ -0,0 +1,44 @@
+// Straight rip off from https://usehooks.com/useEventListener/
+// Left this as js cuz i didn't want to figure out the types.
```
bunnymatic, line 2: this is fine for now.  might be a good exercise to try to type this helper at some point but not required now.

frontend/javascript/react/hooks/use_event_listener.js
```diff
@@ -0,0 +1,44 @@
...
+
+      // Add event listener
+      element.addEventListener(eventName, eventListener);
+
+      // Remove event listener on cleanup
+      return () => {
+        element.removeEventListener(eventName, eventListener);
+      };
+    },
+    [eventName, element] // Re-run if eventName or element changes
```
bunnymatic, line 38: do we not need `savedHandler` in here as a dependency because it's a `useRef`?  I'm not really sure.

frontend/javascript/react/components/art_carousel/art_carousel.tsx
kneyzberg: I see, this does make sense, I'll add the EVENT_KEYS to the useEventListener

frontend/javascript/react/hooks/use_event_listener.js
kneyzberg: Yeah I think that's correct, an update in it's value will not trigger a re-render so we shouldn't add it to the useEffect.

bunnymatic (approved):

PR #233, opened 2024-01-13
Reshaping ArtPiece Model Data for use in Frontend

kneyzberg:
### Problem 
Currently, the data that we were surfacing for the Art Carousel, did not have access to the actual medium of the art Pieces. 

### Solution 
Create an `as_json` method on the ArtPiece model that only includes the pertinent data that we need, and turns into JSON, for easy consumption on the frontend


app/models/art_piece.rb
```diff
@@ -41,6 +41,22 @@ def thumbnail
     images['small']
   end
 
+  def as_json
+    {
+      'artist_name' => @model['artist_name'],
+      'price' => @model['price'],
```
bunnymatic, line 47: since this is acting kind of like a presenter (preparing data for the frontend) do we need `price` and `display_price`?  maybe we just need only one.

bunnymatic, line 47: why do some of these come from `@model` and some are using the attribute methods defined in this class?  maybe could leverage our attribute methods where they exist and only use `@model` if we didn't have that helper?  or... actually write the helpers and use that here.

If we had all the helpers, I could totally see this as
```
{ artist_name: artist_name, price: price, year: year, dimensions: dimensions, ... }.as_json
```

app/models/art_piece.rb
```diff
@@ -41,6 +41,22 @@ def thumbnail
...
+      'display_price' => price,
+      'year' => year,
+      'dimensions' => dimensions,
+      'title' => title,
+      'image_urls' => @model['image_urls'],
+      'sold_at' => @model['sold_at'],
+      'id' => @model['id'],
+      'tag_ids' => @model['tag_ids'],
```
bunnymatic, line 55: we're not using `tag_ids` yet so let's leave it out.  If we do, we'll likely send tags (like we're doing for medium) here instead of ids, so it's probably not going to serve much purpose even if we left it.

PR #235, opened 2024-01-14
Move config.yml so it's not checked in with secrets

bunnymatic:
problem
--------

we need a way to configure the app and also not provide secrets in our
repository (on github or elsewhere) that someone could find.

Also because of the way we deploy, using env vars is not as straight
forward.

solution
--------

let github handle some example `config.yml` files and then load a file
based on environment (or fall back to a `config/config.yml`).

Also, ignore some config files in git so that when we actually deploy
we'll need to manually setup a config file on the deployeed machine
which forces secrets to be kept secret.


kneyzberg (approved):
This all looks great, and sets us up well for deployment 👍

PR #245, opened 2024-01-16
Add capistrano

bunnymatic:
problem
---------

we need to deploy.

solution
--------

this adds `capistrano` as our deploy tool.

After some configuration issues we got this working.

Changes
----------

* add capistrano and helper gems
* add Capfile, and capistrano configuration under config/deploy*
* Add capistrano migration tasks  - cribbed from capistrano/rails
* Add our own version of local precompile tasks - cribbed from
  capistrano-local-precompile and modified for Sinatra (rack_env)
* updated app to not include `byebug` by default (not in production gemset)
* Update readme to talk about how to deploy
* Better ignoring of rspec loading


config/deploy/production.rb
```diff
@@ -0,0 +1,68 @@
...
+#   user: "user_name",
+#   roles: %w{web app},
+#   ssh_options: {
+#     user: "user_name", # overrides user setting above
+#     keys: %w(/home/user_name/.ssh/id_rsa),
+#     forward_agent: false,
+#     auth_methods: %w(publickey password)
+#     # password: "please use keys"
+#   }
+after 'deploy:published', 'deploy:restart'
```
bunnymatic, line 68: I was hoping this would restart `puma` but it seems to need a little work/investigation.

PR #247, opened 2024-01-16
Fix react component styling on deployed server

bunnymatic:
problem
--------

some of our styling wasn't coming through on a deployed server.

solution
--------

we need to included *all* css in our `application.scss` and/or
`admin.scss`.

The problem here is that in development mode, Vite sees all and just
builds all the js and css.

But if we haven't `@use`d the file in our main css entrypoint, when it
gets built out, we miss the boat and those css files for our components
are not included.

Changes
------

add all the react css files to the `application.scss` main file.

Screenshots
--------

**before**

Screen Shot 2024-01-15 at 10 52 07 PM
Screen Shot 2024-01-15 at 10 52 00 PM

**after**
Screen Shot 2024-01-15 at 10 53 14 PM
Screen Shot 2024-01-15 at 10 52 23 PM

Testing
------

Ran `bin/rackup` without `bin/vite dev` so it'd just build assets and
serve them up more like production.

With that I saw the issue, made changes and saw the fixes.

Thoughts
----

Maybe we rethink where css sits for js.  or... do we go full bore
css-in-js for components?


bunnymatic:
Because we got deployment now!!! I deployed this branch.  and it fixed the issues.

kneyzberg (commented):
Ahhh this makes sense, thanks for explaining!

PR #250, opened 2024-01-19
Add a little punch to the modal coloring

bunnymatic:
problem
------

i felt like the carousel was a little gray.

solution
--------

* make it a little bigger
* make it white
* make the background a little darker so it sits forward
* move overlay inline styles for react modal into css
* break art_carousel scss into partials that contain the thing that
  matches their class prefix


kneyzberg (approved):
Looks good, I think it's a good improvement, and I like that it takes more of the page and the background is not as distracting when you scroll.

PR #251, opened 2024-01-19
Instagram API Service and Repo setup

kneyzberg:
### Problem
In order to have better control over the look and styling of the the Instagram feed on the website, we need to make calls to the Instagram API to get the image data. 

### Solution 
- set up Instagram API client to make calls to API
- set up Instagram repo to fetch the data we want from Instagram for 1890 bryant


app/services/instagram_api/api_client.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+require 'faraday'
+
+module InstagramApi
+  class ApiClient
+    def self.get(path, params, headers: {})
```
bunnymatic, line 7: to my points above, i might want the "client" to know about and dictate the paths.  This might not line up with how we're doing MAU models and if not we could align.  But i might prefer something like

```
# in a comsumer

posts = InstagramRepo.posts( params ... if necessary like max number or up to date range or something)

class InstagramRepo
   def self.posts
      Instagram::Client.posts.map {|media| Instagram::Post.new(m) }
   end

   def self.profile # we don't need this but to show how this could expand
      Instagram::Profile.new(Instagram::Client.profile)
   end
end

class Instagram::Client

   def self.posts(params as needed)
      self.get_json(/me/media)
   end

   def self.profile(params as needed)
      self.get_json(/me/profile)
   end

   class << self
     private
        def get_json
        end
        def get ... if we even need this little interim
    end
end

```    

does that make sense?  Not sure if this is *too* much unpacking of all this stuff but in my mind - if we're going with Repo... then it feels to me like the repo should call the client and wrap the data for the consumer.  The client should know how it can talk to the backend (the actual instagram) so the knowledge that "posts" means you call /me/media is in the client, not in the repo.  But the client returns kind of raw ish data which the repo knows how to wrap up.

One could also imagine without the Repo layer and just put it all in the Client so the client knows how to grab the data from the right place and wrap it in a nice object for return.

I kind of like the repo/client separation.  Again might be *slightly* over engineered for our needs but...

consider a change to the instagram api shape or endpoint name... who needs to know about that change and have you wrapped up the logic such that the changes could be very surgical?  with the repo client as i kind of outlined above...

if the paths change (me/media becomes /me/media/v2 or something), one tiny change in the client and everyone's happy.

if the api shape changes, the change would be in the Instagram::Post which knows how to translate the return json shape to our little helper.

in both cases the consumer *and* the repo probably would need no modification.

food for thought.

bunnymatic: I suppose the other thing about the outline i suggest above - we can easily wrap the "private" functions because the client should basically hide any "fetch" logic.  So by putting our fetch helpers `get` and `get_json` in the client as private methods, we need to think about "what is the public api" and it kind of forces the conversation and maybe a move towards a `get_media` or `get_posts` or `posts` kind of method that is the public interface.

app/repos/instagram_repo.rb
```diff
@@ -0,0 +1,17 @@
...
+    def self.get_data
+      conf = BryantStreetStudios::App.settings
+      api_path = '/me/media'
+      fields = %w[id media_type media_url permalink thumbnail_url timestamp].join(',')
+      params = {
+        fields:,
+        access_token: conf.instagram_access_token,
+        limit: 20
+      }
+      InstagramApi::ApiClient.get_json(api_path, params)['data']
```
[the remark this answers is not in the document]
kneyzberg: I see, yeah that makes sense, It's a get request, but we can make it so it's wrapped in something nicer!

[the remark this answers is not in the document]
kneyzberg: Maybe something like Instagram Feed? or Posts? I think either would work

app/services/instagram_api/api_client.rb
kneyzberg: I think that makes sense, I like the idea of it being separate but I also totally get the client being smarter about paths. I think this would be a good way to go

app/models/instagram/instagram.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+module Instagram
+  class Post
+    attr_reader :model
```
bunnymatic, line 5: do we need this?  not a big deal but if we're going to use it like we do in react, maybe better to make an `as_json` method.

app/services/instagram_api/api_client.rb
```diff
@@ -0,0 +1,41 @@
...
+
+module Instagram
+  class Client
+    def self.posts
+      fields = %w[id media_type media_url permalink thumbnail_url timestamp].join(',')
+      req_data = {
+        path: '/me/media',
+        params: {
+          fields:,
+          access_token: BryantStreetStudios::App.settings.instagram_access_token,
```
bunnymatic, line 13: this is true for all requests right? so maybe it should be in the `get_json` definition.

app/services/instagram_api/api_client.rb
```diff
@@ -0,0 +1,41 @@
...
+          limit: 20
```
bunnymatic, line 14: this might be nice as a named constant even if just in this file.  so it's easy to find and change if we decide on 16 or 24 or whatever.

app/services/instagram_api/api_client.rb
```diff
@@ -0,0 +1,41 @@
...
+      }
+      get_json(**req_data)['data']
+    end
+
+    class << self
+      private
+
+      def get_json(path: '', params: {}, headers: {})
+        json_headers = { 'Content-Type' => 'application/json' }.merge(headers)
+        resp = Faraday.get(uri(path)) do |req|
```
bunnymatic, line 25: what happens here if we get a non 200 response?  do we need to catch/rescue something?

bunnymatic (commented):
looking good.

tests?

bunnymatic (commented):
I also wonder if general we should guard against the instagram_key or instagram path being undefined?

maybe get_json could just return {} if those are not there?

spec/repos/instagram_repo_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  before do
+    allow(Instagram::Client).to receive(:get_json).and_return(instagram_data)
+  end
+
+  describe '.posts' do
+    it 'returns the posts' do
+      posts = described_class.posts
+      expect(posts.first).to be_a Instagram::Post
+      expect(posts.first.media_type).to eq 'VIDEO'
```
bunnymatic, line 28: oh boy.  are we going to have to handle video rendering?

spec/services/instagram/instagram_api_client_spec.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+require 'faraday'
+
+describe Instagram::Client do
+  describe '.posts' do
```
bunnymatic, line 7: This is not bad.  But the fact that you're mocking methods on the object you're testing feels a little off.  

Might be better if we looked into using `stub_request`.  

you can see it use in `media_repo_spec.rb`.

it'd be something like
```
stub_request(:get, 'http://instagram.com/me/media').to_return(status: 200, body: ...)
```
and for errors - does the endpoint documentation talk about errors?
```
stub_request(:get, 'http://instagram.com/me/media').to_return(status: 400, body: ...)
```

This kind of helps verify that the Faraday error handling is right because we're stubbing at the network layer kind of just below Faraday.

app/services/instagram_api/api_client.rb
```diff
@@ -0,0 +1,67 @@
...
+            req.headers = json_headers
+          end
+          Oj.load(resp.body)
+        else
+          {}
+        end
+      end
+
+      def base_url
+        config.instagram_api_url
```
bunnymatic, line 45: if you used safe navigation operators here, would that also work?  `config&.instagram_api_url`?  might just be a little cleaner.

app/services/instagram_api/api_client.rb
```diff
@@ -0,0 +1,67 @@
+# frozen_string_literal: true
+
+require 'faraday'
+
+module Instagram
+  class Client
+    MAX_NUMBER_POSTS = 20
+    def self.posts
```
bunnymatic, line 8: this looks really nice and clean!

app/services/instagram_api/api_client.rb
```diff
@@ -0,0 +1,67 @@
...
+      end
+
+      def access_token
+        config.instagram_access_token
+      rescue NoMethodError
+        nil
+      end
+
+      def faraday_connection
+        Faraday.new(url: base_url) do |faraday|
```
bunnymatic, line 57: +1

spec/services/instagram/instagram_api_client_spec.rb
kneyzberg: I see, yeah this makes a lot more sense!
There are some docs on error codes, they are in the 400s
https://developers.facebook.com/docs/instagram-basic-display-api/reference/error-codes

app/services/instagram_api/api_client.rb
kneyzberg: Ahh yess, this was what i was looking for but couldn't figure it out, thanks!

bunnymatic (approved):

bunnymatic:
last note - we should add the new keys to `config.example.yml`

PR #258, opened 2024-01-21
Accessibility Updates

bunnymatic:
problem
-------

we want to try to be accessible!

solution
------

Run a simple accessibility audit using Safari dev tools and correct the
errors.

Primary errors were simply labeling of links.


kneyzberg (approved):
Makes sense, def will make sure to keep this in mind as we continue to do more design revamp stuff

PR #260, opened 2024-01-21
Add helper to compute background image

bunnymatic:
problem
------

we use the background image trick a bunch.  let's wrap it up in a
helper.

solution
--------

steal `backgroundImageStyle` helper from MAU and use it.


frontend/javascript/services/helpers.js
```diff
@@ -1,5 +1,5 @@
 const identity = (x) => x;
-const noop = (_) => {};
```
bunnymatic, line 2: moved these under `services/` directory and lint fix.

kneyzberg (approved):
Awesome, this is super helpful!

PR #265, opened 2024-01-22
Let's move back to artpiece.id

bunnymatic:
problem
------

There we some issues that came up in #233 with `id` not working for some
reason and so we went with `artpiece_id` as the attribute.  We know this
is not ideal since we're already on an art piece.

solution
-------

i flipped it back to use the `MauModel` parent class's id method and
then tighened up the typing on the javascript side.  I did this by
running

```
> art_piece = Repos::ArtistRepo.all.first.art_pieces.first
> art_piece.as_json
```

What I found was that `id` is an integer and year is an integer and
everything else was a string.  So I added this to the frontend typing
and added a small change in our art piece finder to make sure when we're
comparing id's, we are doing apples to apples (integers => integers).


kneyzberg (approved):
Checked this again, and looks good, I think was an issue with the server not running possibly when I was testing.

PR #277, opened 2024-01-23
Revamp artist browser grid to try to be more squareer

bunnymatic:
problem
-------

our artbrowser grid doesn't resize *that* well.  it's not bad but...

solution
-------

experimental - try a little different approach.

wrap the artist browser and make it sized based on page height
for tall browsers.  then force it square.

else (small browsers) do what we used to do basically.

notes
----

kind of weird with a semi-random media rule based on height but... it's
a thing.

screenshots
-------
from big to small
Screen Shot 2024-01-22 at 7 48 11 PM

Screen Shot 2024-01-22 at 7 48 19 PM

Screen Shot 2024-01-22 at 7 48 28 PM


kneyzberg (commented):
Do you have any artist with 2 images? I don't love how it puts the images at the bottom of the grid instead of the top. Are you seeing that?

bunnymatic:
yep - that was kind of by design.  it is centering the whole grid on the right side of the page.  but i agree it's a little weird.

kneyzberg:
I see, I guess we can implement, and see what our designer say, since this won't be the final polish?

PR #278, opened 2024-01-23
Move to lazy load images and revamp artist card grid

bunnymatic:
problem
------

we had some problems with resizing and the logo blowing out the grid.

Also because we were using background images, we couldn't leverage the
image `loading=lazy` tag.

solution
---------

move away from the background image for the list page.  instead put
an image in there and use `object-fit` to scale it.  Then change
the calculation of how to put the image and the name in the box.

this allows for better scaling and means we can leverage the
loading=lazy image attribute.

screenshots
-----------
Screen Shot 2024-01-23 at 1 08 36 PM
Screen Shot 2024-01-23 at 1 08 48 PM
Screen Shot 2024-01-23 at 1 08 56 PM
Screen Shot 2024-01-23 at 1 09 11 PM


kneyzberg (approved):
Love it, I think it's awesome!

PR #279, opened 2024-01-23
Slight rearrange of app settings

bunnymatic:
problem
-------

app.rb was kind of messy

solution
-----

in looking to maybe a safe settings fetcher, i decided to clean up the
code sitting in app a little bit.

organize all the setting setters
don't duplicate setting vars that are already in settings


app.rb
```diff
@@ -7,10 +7,6 @@
 require 'sinatra/static_assets'
 require 'padrino-helpers'
 
-require 'oj'
-require 'uri'
```
bunnymatic, line 11: moved `oj` requirement to where we need it.
others are not required here.

app.rb
```diff
@@ -72,12 +69,6 @@ class App < Sinatra::Base
 
     config_file ConfigFileFinder.find(environment, root)
 
-    # fetch from env or config
-    set :auth_user, settings.auth_user
```
bunnymatic, line 76:
```
set :attr, settings.attr
```
is repetitive.  if it's in `settings` it was already set.  These were left over from when we looked at the unix env vars.

PR #280, opened 2024-01-24
Add Mau::Client layer

bunnymatic:
problem
------

after discussing API strategies and where our `repo` lives and where
our clients live, we agreed that the client layer's we've built for MAU
are kind of at the wrong level of responsibility.

In looking to add a better way to handle updating the artist list which
we need, and how to clear the cache for that, this discrepancy became
more visible.

solution
---------

this PR introduces a Mau::Client which has all the fetcher methods that
we need.  It and only it calls the underlying `Mau::BaseJsonApiClient`
and `Mau::BaseApiClient` (renamed to add `Base` to indicate - you
probably shouldn't reach into these from outside).

There are still some area's where cache control might be leaking across
services but this is closer and should allow us to update the
`get_1890_bryant` method to correctly use or not use a cache which will
allow more surgical cache management as we add webhook methods that
might want to  just update the list of artists when a new artist is
added on MAU.

Notes
----

In this revamp, the shape of the API calls changed a bit which means a
bunch of our fixtures needed adjustment.

We were calling `api/v2//artists/` and the update fixes this path to be
`api/v2/artists/`.  I did *not* re-record but just manually updated the
fixtures.


app/services/mau/base_api_client.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+module Mau
+  class BaseApiClient
+    def self.get_json(path, headers: {})
```
bunnymatic, line 7: pretty much a straight move from `ApiClient`.  not sure why git didn't call it a move.

app/services/mau/client.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+module Mau
+  class Client
```
bunnymatic, line 4: this class now wraps up (as we've talked about) the data we can get from mau with methods that indicate that data.  it hides the paths required and any other call tricks (like taking `['data']` off the top) in here so callers can be simpler.

It also pushes the cache control down.  Though we do indicate to callers 'there is a cache' by including a `cache` argument on some of these methods, the caller doesn't need to know *how* to clear the cache.

more will come as we figure out how to handle updating the artists id list and trying to allow for updating the artist list without clearing the entire cache.

app/services/mau/client.rb
```diff
@@ -0,0 +1,36 @@
...
+      def get_artist(id, cache: true)
+        (Mau::BaseJsonApiClient.new(cache:).get_json("artists/#{id}.json") || {})['data'] || {}
+      end
+
+      def get_art_pieces(artist_id, cache: true)
+        (Mau::BaseJsonApiClient.new(cache:).get_json(art_pieces_path(artist_id)) || {})['data'] || []
+      end
+
+      def get_media
+        BaseApiClient.get_json("#{config.mau_api_prefix}/media.json")
```
bunnymatic, line 22: Seeing this makes me think maybe the distinction between the json_api client and the api client is silly and if it was all in the `BaseClient` maybe things would be simpler.

Someone has to know (maybe here) that the data from some endpoints is the `jsonapi` format and some is not.  I guess in the same place we specify the path, is the place where we could/should have that information.  so...  maybe a follow up PR.

I think It would further limit the caching logic since currently the JsonApiClient caches each request but the BaseApiClient.  

We have discussed and wondered if having the request level cache is necessary or if it's just over complicating things.  so... if we put it all in one place, we could more easily dump it or not

app/repos/artist_repo.rb
```diff
@@ -46,8 +50,8 @@ def self.find_neighbors(id)
 
     def self.refresh_artist(id)
       SafeCache.delete(artist_cache_key(id))
```
bunnymatic, line 52: I still don't totally like how much this thing knows but it feels one layer better maybe.

app/services/mau/base_json_api_client.rb
```diff
@@ -10,18 +10,14 @@ def get_json(path)
       resp = @cache ? SafeCache.get(path) : nil
       return resp if resp.present?
 
-      Mau::ApiClient.get_json(api_url(path)).tap do |data|
-        @cache && SafeCache.set(path, data)
+      Mau::BaseApiClient.get_json(api_url(path)).tap do |data|
```
bunnymatic, line 13: change this logic to *always* cache when we have data but if `cache: false` we don't read from the cache.

PR #281, opened 2024-01-25
Collapse MAU base api clients

bunnymatic:
problem
------

After #280 it felt like the 2 mau api clients could be collapsed into
one.

solution
------

sure enough... it was easy.

* combine/collapse `Mau::BaseJsonApiClient` and `Mau::BaseApiClient`
* move request caching down to the `get`
* move api client tests to use `stub_request` to better handle errors
* improve error handling in base client


kneyzberg:
Awesome, love the simplicity now!

PR #282, opened 2024-01-25
Add `.from_jsonapi` method to favor over `.new` for models

bunnymatic:
problem
--------

our models should be a teeny bit dumber - or maybe just more explicit.

solution
------

Add a `from_jsonapi` helper initializer which calls `new` with a simple
model hash that has been deserialized from jsonapi.


app/models/mau_model.rb
```diff
@@ -3,10 +3,14 @@
...
-  include JSONAPI::Deserialization
+  extend JSONAPI::Deserialization
 
-  def initialize(json_api_data)
-    @model = jsonapi_deserialize(json_api_data)
+  def self.from_jsonapi(jsonapi_data)
+    new(jsonapi_deserialize(jsonapi_data))
+  end
+
+  def initialize(data)
```
bunnymatic, line 12: with this change, we could make our factories make simpler data and not necessarily have to know all about jsonapi format.  

even without that, it makes it clear that we're initializing this model with jsonapi.  so if we wanted something else, we could add a different `from_other_format` initializer.

PR #306, opened 2024-02-02
Create Styled Instagram Feed

kneyzberg:
### Problem
The current instagram embed has styling that we are unable to modify or control. 

### Solution 
Build our own Instagram Feed using the Instagram Basic Display API and display it on frontend with React. 

- Created InstagramFeed and InstagramCard component with styling 
- rendered test grid on admin page
- Updated Icomoon icons to include stack and play button
- Implemented frontend api service for fetching the Instagram Data
- Added tests 
Screen Shot 2024-02-18 at 10 27 54 PM


app/views/admin/test/index.slim
```diff
@@ -53,3 +56,8 @@ ruby:
     h1 ArtCard
     .react_component
       = react_component(id: 'artCard', component: 'ArtCard', props: {art_piece: art_pieces[0]})
+  
+  article.components-index
+    h1 InstagramFeed
+    .react_component
+      = react_component(id: 'InstagramFeed', component: 'InstagramFeed', props: {posts: instagram_data})
```
bunnymatic, line 63: this immediately makes me think we should have a proxy api endpoint so that the react component can call bss for the data, instead of waiting for the server to pull together all artists *and* all instagram.  it gives us the option for spinners and parallel requests etc.

this is a fine place to start but before we deploy with instagram we should probably think about some ajax work in react.

frontend/javascript/react/components/instagram_feed/instagram_card.scss
```diff
@@ -0,0 +1,21 @@
+@use '../../../../stylesheets/layout.scss';
+
+
+.instagram-card {
+
+  height: 30vw;
```
bunnymatic, line 6: I wonder if some if this crazy height/width stuff can go away for *this* because we maybe know that every instagram image is already sized to something?  that is, if they are alway square, maybe this can be simplified in some way.  not sure.  just a thought.

bunnymatic (dismissed):

bunnymatic:
totally going the right way

frontend/javascript/react/components/instagram_feed/instagram_feed.tsx
```diff
@@ -0,0 +1,56 @@
...
+const InstagramFeed: FC = () => {
+    const [data, setData] = useState<InstagramFeed>()
+    const [loading, setLoading] = useState<boolean>(false)
+
+  const fetchInstagramData = useCallback(async () => {
+    try {
+      setLoading(true)
+      const response =  await api.instagram.feed()
+      setLoading(false)
+      console.log(response, "FETCH")
```
bunnymatic, line 24: 💣

frontend/javascript/react/components/instagram_feed/instagram_card.tsx
```diff
@@ -0,0 +1,35 @@
...
+    ? instagramPost.thumbnailUrl
+    : instagramPost.mediaUrl;
+  const imageStyle = backgroundImageStyle(imageUrl);
+
+  return (
+    <div className={cx("instagram-card", classes)}>
+      <a
+        href={instagramPost.permalink}
+        // onClick={handleClick}
+        // title={artPiece.title}
```
bunnymatic, line 26: do we have any good content for the title? doesn't look like it based on the typescript shape of the response post. but if we could grab anything like the  post comment or something, might be nice for accessibility.

frontend/javascript/react/components/instagram_feed/instagram_feed.scss
```diff
@@ -0,0 +1,19 @@
+@use '../../../../stylesheets/layout.scss';
+
+
+.instagram-feed--grid {
+  display: grid;
+  flex: 1;
+  gap: 10px;
+  grid-template-columns: repeat(2, 1fr);
+
+  @media screen and (min-width: layout.$screen-md-min) {
```
bunnymatic, line 10: can grid have justify and align?  just as a precaution, might be good to add `justify-content: cetner/align-items: center` if the grid honors that in the cells.

bunnymatic (commented):
looks good.  how about a test of the instagram component?

frontend/javascript/react/components/instagram_feed/instagram_card.tsx
kneyzberg: I'll take a look, but i don't think we get that data back, but let me confirm

frontend/javascript/react/components/instagram_feed/instagram_feed.scss
kneyzberg: sounds good, will add that/fix

kneyzberg:
> looks good. how about a test of the instagram component?

Yup definitely, adding that in, this was just me rebasing everything to get the latest in the branch, nothing new here yet from when we worked on it last!

frontend/javascript/react/components/instagram_feed/instagram_card.tsx
kneyzberg: Yeah we can get the caption for the posts, I can add that in

frontend/javascript/react/components/instagram_feed/instagram_card.tsx
```diff
@@ -0,0 +1,44 @@
...
+    ? instagramPost.thumbnailUrl
+    : instagramPost.mediaUrl;
+  const imageStyle = backgroundImageStyle(imageUrl);
+
+  return (
+    <div className={cx("instagram-card", classes)}>
+      <a href={instagramPost.permalink} title={instagramPost.caption}>
+        <div className="instagram-card--image" style={imageStyle}></div>
+        {instagramPost.mediaType == "CAROUSEL_ALBUM" && (
+          <div
```
bunnymatic, line 27: these feel kind of repetitive and complex where the only real thing that is changing is the icon name, right?  and `testid` for RTL.  

I wonder we could rejigger it and still get a good test by doing something like


```
let iconName;
switch(instagramPost.mediaType) {
  case "CAROUSEL_ALBUM":
     iconName = 'stack'
  case "VIDEO":
     iconName = 'video'
  }

  <div><icon name={iconName}/></div>
```

frontend/javascript/react/components/types/instagram_post.ts
```diff
@@ -0,0 +1,11 @@
+interface InstagramPost {
+  id: string;
+  mediaType: string;
```
bunnymatic, line 3: do we know enough to type this like
"VIDEO" | "IMAGE" | "WHATEVER"?  If not, can we make some constants somewhere so that the loose comparison above to =="CAROUSEL" has some typescript guards.

frontend/javascript/services/api.ts
```diff
@@ -10,7 +12,9 @@ const request = {
 const api = {
   instagram: {
     feed: () => {
-      return request.get("/api/instagram/feed").then(camelize);
+      return request
+        .get("/api/instagram/feed")
+        .then((posts) => camelize(posts) as types.InstagramPost[]);
```
bunnymatic, line 17: Does this feed function get a return type like `: Promise<InstagramPost[]>`?  maybe?

bunnymatic (approved):
Looking really good.  We can likely merge it.  I just wonder if we want to take the typescript stuff just a teeny bit further.  Honestly, I'm a little on the fence.

frontend/javascript/react/components/instagram_feed/instagram_card.tsx
kneyzberg: oh yeah, forgot about switch cases this is wayyyy better

frontend/javascript/react/components/types/instagram_post.ts
kneyzberg: yeah we know if will be one of three things, can totally add that in

frontend/javascript/services/api.ts
kneyzberg: maybe that would be better, i was having major issues typing this, but worth exploring

PR #307, opened 2024-02-04
Replace memcache cache with database records

bunnymatic:
problem
---------

to properly cache *and* make it easy to figure out what needs a refresh
without a lot of work, the memcache solution doesn't really work.  A
database is a better choice.

The main issue is related to the API.  We can either ask for *all*
artists and get most of what we need in 1 payload, or we can make 100's
of calls.  And knowing what to call is pretty much down to the artist
ids.  Because it's hard to unpack the id's list and insert entries if we
found that an artist got added or deleted is trickier with memcache, and
clearing the cache meant a bunch of load on MAU, we decided for this
approach.

solution
-------

Instead of caching with memcache, we do the following:

* if we have no records in our db, ask MAU for the full payload and then
  with each record, added it to the db.  Also cache the id's we have.
* for each load after that, just fetch each artist by id.  If we have
  the artist, we're done.  If not, pull from MAU and stick it in the db.
* refresh now is as simple as deleting the database record and
  refetching from MAU.

changes
--------

* move Artist model to Entities::Artist
* update Entities::Artist to be a data model that could be initialized
  by a database record *or* the raw jsonapi data
* some puma refactors because I found tests were hanging up
* stop caching as aggressively at the request level
* update database cleaner to run on truncation to avoid data leaking
  across specs
* add webhook integration to prove things out
* move to using `mau_artist_id` over `id`.  Our local db id doesn't mean
  much since almost everything is driven by the id from MAU.
* re-record and update specs
* better filtering of artists without art
* rename a bunch of tests and shuffle fixtures to get tests up to date


app/repos/artist_repo.rb
```diff
@@ -29,76 +29,110 @@ def flush
...
-      # if not, call all artists
-      #    pluck ids and cache them
-      #    cache artist data from that call
-      #
-      # when new artist is created, populate artist id cache from artist/ids call
-      #    next time new artist
-      artist_list = artist_list_from_cached_artist_ids || artist_list_from_client_get_all
-
-      (include_exclusions ? artist_list : filter_excluded_artists(artist_list)).sort_by(&:last_name)
+      artists = if Artist.count.zero?
```
bunnymatic, line 122: If we have *no* data in the db, let's make 1 call.  else fetch 1 by 1.

this is kind of dumb - it is pushing knowledge about our 3rd party behavior and worrying about it's performance here on the consumer.  But... since we *do* know this,   this is smarter.

app/services/mau/base_api_client.rb
```diff
@@ -7,11 +7,11 @@ class BaseApiClient
     ERROR_CODE_RANGE = 400...510
     class Error < StandardError; end
 
-    def self.get_json(path, cache: true)
+    def self.get_json(path, cache: false)
```
bunnymatic, line 10: this is to be a little less cache aggressive

app/services/mau/client.rb
```diff
@@ -24,7 +24,7 @@ def get_art_pieces(artist_id, cache: true)
       end
 
       def get_media
-        BaseApiClient.get_json('media.json')
+        BaseApiClient.get_json('media.json', cache: true)
```
bunnymatic, line 27: medium list basically never changes so...

config/puma.rb
```diff
@@ -0,0 +1,4 @@
+# frozen_string_literal: true
+
```
bunnymatic, line 2: I don't know what changed here but puma started not shutting down with integration tests.  This is to force puma to run a single worker process.  We should be careful with deployment to make sure the production server config is separate from this.

spec/controllers/webhook_controller_spec.rb
```diff
@@ -2,25 +2,119 @@
 
 require 'spec_helper'
 
-describe BryantStreetStudios::App, type: :feature do
-  before do
-    allow(Repos::ArtistRepo).to receive(:refresh_artist)
-  end
+describe 'Webhook Controller', type: :feature do
```
bunnymatic, line 5: I renamed a bunch of these test suites to indicate what we're testing.

And that mean moving a bunch of fixture files to new directories.

spec/controllers/webhook_controller_spec.rb
```diff
@@ -2,25 +2,119 @@
...
     end
+
+    it 'refreshes the artist ids list' do
+      post '/webhook/artist/7/update', {}, { 'HTTP_BSS_API_AUTHORIZATION' => conf.webhook_auth_key }
+      expect(Repos::ArtistRepo).to have_received(:refresh_artist_list)
+    end
+  end
+
+  describe 'integration spec' do
+    let(:jon) { create(:artist, full_name: 'jon') }
```
bunnymatic, line 36: I wanted to build a full flow test.


Start with 1 artist
update the first artist
verify that we get the updated data
pretend 1 new artist was added to mau
make sure they show up here
delete an artist
make sure the new list of artists is correct


it's not a pretty test but it did prove out the behavior.

spec/support/capybara.rb
```diff
@@ -45,12 +45,21 @@ def self.chrome_capabilities(headless: false)
 # for screenshot/html saving
 Capybara.save_path = 'tmp/capybara'
 
+Capybara.register_server :puma do |app, port, host|
+  require 'rack/handler/puma'
```
bunnymatic, line 49: as noted above, this register server seems to square up the `puma` single worker and allow it to shut down at the end.

this and the `at_exit` below.

PR #323, opened 2024-02-11
undo css grid to rearrange content on mobile

bunnymatic:
problem
-----

with #286 we introduced the css grid with named template areas with the
idea that we could more easily rearrange content for mobile.

It seemed great at the time but the grid imposes height issues on the
left column that we did not forsee.

In a spike to not draw the carousel when there was only one image, I ran
into layout troubles because the grid heights of things in the left were
affecting the available height on the right making it very hard to
simply draw a full size single image.

solution
--------

chuck the grid in favor of some more simple flex tricks and simply
rendering the bio content twice on the page in the dom elements where
wanted it.  The bio now is rendered with classes that show and hide the
appropriate version depending on the browser size.


bunnymatic:
@kneyzberg sorry to rip this out after we did a bunch of work this direction but it was really getting in the way.

kneyzberg:
No worries! With flexbox it's probably easier to change things!

PR #326, opened 2024-02-12
Render art with details if there is only 1 piece

bunnymatic:
problem
------

the carousel seems a little crazy when there is only 1 piece of art.

solution
--------

render the art if there is only 1 piece.

Screenshots
----------

1 piece of art
Screen Shot 2024-02-11 at 4 20 26 PM

multiple pieces - still have the carousel

Screen Shot 2024-02-11 at 4 20 34 PM


frontend/javascript/react/components/art_piece_details/art_piece_details.scss
```diff
@@ -11,10 +11,10 @@
 }
 
 .art-piece-details--right {
-  text-align: right;
```
bunnymatic, line 14: flipped these rules so we're mobile first.
then adjusted where we move right to the right

PR #333, opened 2024-02-13
Create Internal API for Instagram feed 

kneyzberg:
### Problem 
As more and more of our frontend is written in React, we want the ability to make calls to the backend  so we have more control over what we show on the page while we are retrieving data. 

### Solution
Instead of having the server make the call to retrieve the Instagram data before rendering the page, create an internal API that the React Frontend can call to retrieve the Instagram Data. 

- Created an Instagram API controller to return json data 
- Created an api service on the frontend to make an AJAX request to the backend. 
- Added tests for controller and frontend api service


frontend/javascript/services/api.test.ts
```diff
@@ -0,0 +1,32 @@
+/**
+ * @jest-environment node
```
bunnymatic, line 2: what is this doing?  i've never seen it.  so curious?

frontend/javascript/services/api.ts
```diff
@@ -0,0 +1,19 @@
...
+const camelize = (data: JsonResponseData) => camelizeKeys(data)
+
+const request =  {
+  get: (path: string) => fetch(path).then(data => data.json())
+}
+
+const api = {
+  instagram: { 
+    feed: () => {
+      return request.get('/api/instagram/feed').then(camelize)
```
bunnymatic, line 13: so clean and simple :D

spec/controllers/api/instagram_controller_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+      'timestamp' => '2024-01-16T20:22:13+0000'
+    }]
+  end
+
+  describe '/api/instagram/feed' do
+    before do
+      allow(Repos::InstagramRepo).to receive(:posts).and_return(instagram_posts)
+    end
+
+    it 'returns Instagram data as JSON' do
```
bunnymatic, line 24: do we know what happens if there is an error?
we could leave that til later or... 
I think we might want to make sure this doesn't hit our error page because these api's are json and should only return json.  

first easy cut would be something like

```
describe 'when there is an error' do
   before do
      allow(Repos::InstagramRepo).to receive(:posts).and raise_error('crap')
   end

   it 'returns nothing' do
     get '/api/instagram/feed'
     expect(last_response.body).to eq('{}')
     # not sure about status but probably should be a 404 NotFound, right?  I couldn't find instagram?  something like that.
   end
```

if you can make a test like that pass, then we've got a little guard.

bunnymatic (approved):

spec/controllers/api/instagram_controller_spec.rb
bunnymatic: I approved.   we can merge this and work on error handling in subsequent work since there is more to stack on top of this - if you want.

it might also be something generic for all our "API" controller methods.  we could maybe leverage some kind of `after` block (not sure how that works in sinatra for a set of methods but... i think you can do something like
```
after %r{/api/*} do
   ... handle errors appropriately
end
```
maybe

frontend/javascript/services/api.test.ts
kneyzberg: It's just so we can mock fetch, you can only do it with running in a node test environment for some reason. 
https://github.com/jestjs/jest/issues/13834#issuecomment-1407375787

https://stackoverflow.com/a/76679669

spec/controllers/api/instagram_controller_spec.rb
kneyzberg: Definitely worth looking into. I think we will need to create a guard in our InstagramRepo.posts function because right now it will make a post with attributes that are null if the client returns {} , so we should def take care of that edge case, and I think it would be worth fixing here as well

bunnymatic: for this api, we can also be pretty forgiving.  if there is a problem just log it maybe and return no data.  but it would be nice to have an example of how we catch api errors and handle them.

kneyzberg: Copy, i'll merge for now so you can build on top of this,  but work on error handling in a separate PR

PR #335, opened 2024-02-13
Move artist page to react

bunnymatic:
problem
-------

we are moving to using React a little more so we can leverage things
like fast page loads and delayed data loading with Ajax and apis.

to get there, we need to get rid of some of our slim templates.

solution
-------

move most of the artist page to slim

changes
------
* add ArtistInfo react component which replaces _artist_info.slim
* add ArtistSocial react component which replaces _social.slim
* update biobox to handle the markup of markdown and test it
* add Artist type and artistFactory for tests
* update artist `as_json` method to include art_pieces the controller
  can do less work

screenshots
---------------
Screen Shot 2024-02-13 at 11 09 31 AM
Screen Shot 2024-02-13 at 11 09 40 AM
Screen Shot 2024-02-13 at 11 11 05 AM


bunnymatic:
needs work - I might have busted tablet ish size layouts along the way here.

image

bunnymatic:
Actually it looks like that was a fail from a previous merge cuz it looks like that on the deployed site so... maybe a bug to fix that separately.  probably from #323 https://www.pivotaltracker.com/story/show/187042637

kneyzberg:
looking awesome!

PR #336, opened 2024-02-14
Fix squished art view

bunnymatic:
problem
------

tablet size was squishing out the art.  It seems the grid with the
profile image on the right was taking up too much space in the tablet
mode which lead to the `flex:1` for the art not getting all the space.

solution
-------

this pr flips the styles to use the grid *only* when we're on the
smallest of screens.  otherwise just flex column on the left artist-info
column.  Also change the flex so that we get a little less than 44%/56%
for tablets then 25%/75% and 20%/80% split as we move to bigger screens

we also flipped a few of the media rules to start mobile first and then
define refinements for screens as they go up in size.

Screenshots
---------

**BEFORE**
Screenshot 2024-02-13 at 4 00 47 PM

**AFTER**

image


kneyzberg:
Frustrating that we can't get away with using either/or but at least this gives us the look that we want

PR #361, opened 2024-02-22
bug/update zindex of artist card 

kneyzberg:
### Problem
On certain browsers, such as chrome, the artist name is stacking above the nav when you scroll down the page, even though the z-index of the nav and the artist card are the same. 

### Solution
Update the z-index if the artist card details so that it falls under the nav when you scroll. 

BUG:
Screen Shot 2024-02-21 at 9 29 29 PM


frontend/stylesheets/zindex.scss
```diff
@@ -7,7 +7,7 @@ $react-bio: 10;
 $ig-feed-icon: 10;
 $editable-content-edit-link: 15;
 $artpiece-card-expand-icon: 19;
-$artist-card-name: 20;
+$artist-card-name: 19;
 $carousel-buttons: 20;
 $mobile-menu: 20;
```
bunnymatic, line 12: I agree on merging this but I wonder if we should just push $nav and $mobile-menu up to 25?

kneyzberg: yeah definitely!

PR #381, opened 2024-03-03
Move react pages to `/pages`

bunnymatic:
problem
------

as we move more to react, let's do a little re-org

solution
------

* add /react/pages/
* add /react/artists/show to hold the artist page react component


bunnymatic:
renamed branch and opened a new pr

PR #383, opened 2024-03-04
Add alias for js imports

bunnymatic:
problem
-----

we have lots of large directory traversing for imports.

solution
-----

for sanity and to make things easier to shuffle around, let's add aliases.

Now we can

```javascript

import { whatever } from "@hooks/whatever";
import { whatever } from "@services/whatever";
import { whatever } from "@components/whatever";
import { whatever } from "@test/whatever";

```

changes
-----

* add paths to tsconfig
* add plugin to read those paths and translate to vite aliases
* add moduleNameMapper for jest to interpret those aliases
* touched every file that had a "../" or more import and tried to replace them all with aliases
* lint-fix to sort imports


bunnymatic:
I purposely did not use aliases on `types` because I'd like to move it up out of `react` since it's kind of a higher level thing.

will have another PR for that.

PR #407, opened 2024-03-15
Move artists page and home page artists list to react

bunnymatic:
problem
-----

in preparation for search, we need to get more of our artist grid stuff into react.

solution
-----

make pages (even if they're not the *full* page) that will render a grid of react
artists and allow for some extra functionality to be jammed in there.

To do this we add `ArtistsPage` and `HomePageArtists` which both use a new `ArtistsContext` provider
that will hold the list of artists and the `ArtistGrid` now consumes artists from that list.

As we move towards search or load artists by json, we can leverage this new context and
pages using it should be able to render as they do now.

changes
-------

* add ArtistsProvider and `useArtists` hook to provide the context data
* wrap the ArtistsGrid with this provider on new home and artist pages
* add `@pages` and `@contexts` import aliases


frontend/javascript/react/contexts/artists_context.tsx
```diff
@@ -0,0 +1,26 @@
...
+  children,
+}) => {
+  return (
+    <ArtistsContext.Provider value={{ artists }}>
+      {children}
+    </ArtistsContext.Provider>
+  );
+};
+
+export const useArtists: () => types.ArtistsContext = () =>
```
bunnymatic, line 25: I put this here instead of a separate hook file.  I feel like a new file under hooks is too much for this kind of thing.  This way the entire context and provider thing are all wrapped up together.

frontend/javascript/react/pages/home/artists/home_page_artists.tsx
```diff
@@ -0,0 +1,17 @@
+import { ArtistsGrid } from "@components/artists_grid/artists_grid";
+import { ArtistsProvider } from "@contexts/artists_context";
+import * as types from "@root/types";
+import React, { FC } from "react";
+
+interface HomePageArtistsProps {
+  artists?: types.Artist[];
+}
+const HomePageArtists: FC<HomePageArtistsProps> = ({ artists }) => {
```
bunnymatic, line 9: this is an exact copy of the `ArtistsPage` but I expect as we move forward these will start to diverge so this sets us up for that.

frontend/javascript/react/pages/artists/index/artists_page.tsx
```diff
@@ -0,0 +1,17 @@
...
+import * as types from "@root/types";
+import React, { FC } from "react";
+
+interface ArtistsPageProps {
+  artists?: types.Artist[];
+}
+const ArtistsPage: FC<ArtistsPageProps> = ({ artists }) => {
+  return (
+    <ArtistsProvider artists={artists || []}>
+      <ArtistsGrid includeNames={true} />
```
bunnymatic, line 12: I'm not sure what we'll do once we're pulling from Ajax but I think this provider should still work nicely in that case so... 

the first of the contexts.

PR #423, opened 2024-03-22
Load Arist Page with JSON

kneyzberg:
### Problem 
Currently, our server server all the artist data to the frontend for the `artist/show` pages. It would be better to have the React component make an API call to the backend, so that we have more control over what we render on the page, such as loading spinners. 

### Solution 
Create an internal api to retrieve Artist Data and implement on the ArtistPage React Component. 
- Create artist controller with api endpoint
- Update services on frontend to make call to backend api endpoint
- Update props for Artist Page component to only be id_or_slug
- Add tests and lint fixes


frontend/javascript/react/pages/artists/show/artist_page.tsx
```diff
@@ -1,37 +1,78 @@
...
 interface ArtistPageProps {
-  artist: types.Artist;
+  artistIdOrSlug: string;
 }
 
-const ArtistPage: FC<ArtistPageProps> = ({ artist }) => {
-  if (!artist) {
+type Artist = types.Artist;
+
+const getCsrfTokenFromMeta = (): string | null => {
```
bunnymatic, line 16: I wonder about making this a helper or maybe a hook.  it's kind of weird the way we've (I've) built it.  not for this PR but it feels like it should be a thing we could just ask for.

I thought about putting this deeper but it's really the responsibility of the DOM so building it down in the api feels wrong.

just sharing thoughts.  no change needed here.  we can do a follow up PR to extract it maybe.

frontend/javascript/react/pages/artists/show/artist_page.tsx
```diff
@@ -1,37 +1,78 @@
...
-        <ArtistArt artPieces={artist.artPieces} />
-      </section>
-      {artist.bio && artist.bio.length ? (
-        <section className="artist-page--bio--mobile">
-          <BioBox bioText={artist.bio} />
+    data && (
+      <div className="artist-page">
+        <section className="artist-page--info">
+          <ArtistInfo artist={data} />
+          {data.bio && data.bio.length ? (
```
bunnymatic, line 60: minor.  could setup a 
```
const hasBio = useMemo(() => data.bio && data.bio.length)
```
and then use that in these two blocks to decide to add (or not) the bio component.

spec/controllers/api/artist_controller_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+  let(:artist) { Entities::Artist.from_active_record(create(:artist, full_name: 'Test Artist', id: 24)) }
+
+  describe '/api/artists/:id_or_slug' do
+    before do
+      allow(CsrfChecker).to receive(:valid?).and_return(true)
+      allow(Repos::ArtistRepo).to receive(:find).and_return(artist)
+      allow(Repos::ArtPieceRepo).to receive(:find).and_return([instance_double(ArtPiece, image: 'a.png')])
+    end
+
+    it 'returns Artist data as JSON' do
```
bunnymatic, line 16: maybe another test (almost duplicate test) that proves that the slug works?

or maybe not since we're mocking stuff here.  maybe it doesn't really mean anything.  IRL do we actually use the slug?  if not, let's make the variable not called id_or_slug, but if we do, then it's fine.  

or maybe a system test will prove this out?  

since you didn't change any system specs, then maybe we're covered.

bunnymatic:
We should think about a CenteredLoader component maybe.  the current one doesn't really fill out the page and if you actually saw it it looks goofy.  again - not for this PR.  but as a note for later


https://github.com/bunnymatic/bryantstreetstudios/assets/427380/9444750b-22f5-46c7-8998-4ca728ef6f0d

bunnymatic (approved):

